PR #7, opened 2015-07-07
Data persistence

sim590:
This PR introduces changes to handle data persistence on the DHT. ~~This is not yet tested, there may have bugs and~~ there are some features that may be modified and/or enhanced. In short : **WIP**.

For now, it means to behave like this :

- all data is "persistent" on the dht until it's expiration time;
- data maintenance is done every 10 minutes for each storage.

@aberaud: Please review this and make comments of any kind!

### TODO
- [x] **preserve value timeout**: For now, all values that are maintained before shutdown have their timeout field reset each time they're sent to another hosting node. Persistent values should be sent with value time stamp information. This may be altering OpenDHT's protocol.
- [ ] ~~**Persistent vs non-persistent values**: there should be a way to make values persistent and others not.~~ Actually, this is not necessary. Preserving the value timeouts on storage maintenance is sufficient behavior.

### DIFF
--- a/include/opendht/dht.h
+++ b/include/opendht/dht.h
@@ -133,6 +133,7 @@ public:
 
     typedef std::function<bool(const std::vector<std::shared_ptr<Value>>& values)> GetCallback;
     typedef std::function<bool(std::shared_ptr<Value> value)> GetCallbackSimple;
+    typedef std::function<void()> ShutdownCallback;
 
     typedef bool (*GetCallbackRaw)(std::shared_ptr<Value>, void *user_data);
 
@@ -156,9 +157,14 @@ public:
 
     typedef std::function<void(bool success, const std::vector<std::shared_ptr<Node>>& nodes)> DoneCallback;
     typedef void (*DoneCallbackRaw)(bool, std::vector<std::shared_ptr<Node>>*, void *user_data);
+    typedef void (*ShutdownCallbackRaw)(void *user_data);
 
     typedef std::function<void(bool success)> DoneCallbackSimple;
 
+    static ShutdownCallback
+    bindShutdownCb(ShutdownCallbackRaw shutdown_cb_raw, void* user_data) {
+        return [=]() { shutdown_cb_raw(user_data); };
+    }
     static DoneCallback
     bindDoneCb(DoneCallbackSimple donecb) {
         if (not donecb) return {};
@@ -197,6 +203,11 @@ public:
     Status getStatus() const {
         return std::max(getStatus(AF_INET), getStatus(AF_INET6));
     }
+
+    /**
+     * Performs final operations before quitting.
+     */
+    void shutdown(ShutdownCallback cb);
 
     /**
      * Returns true if the node is running (have access to an open socket).
--- a/include/opendht/dht.h
+++ b/include/opendht/dht.h
@@ -242,6 +242,13 @@ public:
 
     int pingNode(const sockaddr*, socklen_t);
 
+    /**
+     * Maintains the store. For each storage, if values don't belong there
+     * anymore because this node is too far from the target, values are sent to
+     * the appropriate nodes.
+     */
+    void maintainStore(bool force=false);
+
     time_point periodic(const uint8_t *buf, size_t buflen, const sockaddr *from, socklen_t fromlen);
 
     /**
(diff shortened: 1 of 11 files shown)


src/dht.cpp
```diff
@@ -2120,6 +2120,42 @@ Dht::bucketMaintenance(RoutingTable& list)
 }
 
 void
+Dht::maintainStorage(InfoHash id) {
+    auto *local_storage = findStorage(id);
+    std::vector<std::shared_ptr<Value>> foreign_values;
+
+    if (!local_storage) { return; }
+    get(id,
+    [&foreign_values](std::shared_ptr<Value> value){
```
aberaud: capture by reference of a stack variable

src/dht.cpp
```diff
@@ -2120,6 +2120,42 @@ Dht::bucketMaintenance(RoutingTable& list)
...
+        foreign_values.push_back(value);
```
aberaud: here foreign_values is a dangling pointer

src/dht.cpp
```diff
@@ -2120,6 +2120,42 @@ Dht::bucketMaintenance(RoutingTable& list)
...
+        return true;
+    },
+    [=,&foreign_values](bool success, const std::vector<std::shared_ptr<Node>> &nodes) {
+        if (success) {
+            auto this_node = std::find_if(nodes.begin(), nodes.end(), [this](const std::shared_ptr<Node>& node) {
+                        return getNodeId() == node->id;
+                    });
+            //If this node is not in the list of 8 nodes nearby the given id.
```
aberaud: this node will never be in the list (our own node is not part of routing tables). Instead we should check if we are XOR-closer to "id" than the XOR-further node in the list.

sim590:
I've made some changes and added a benchmark test for the feature. It still needs to be debugged, but it's coming together. What has changed:

- data persistance maintenance is also made before a node quits if the new `shutdown(callback)` call is used.

sim590:
I have worked on solving bugs I had and it now seems right. @aberaud, you can test it using `benchmark.py` script like this:
```
 sudo python3 benchmark.py --data-persistence
```
Using `-n 256` gives better results as for the number of nodes holding persistent values after shutdown of other nodes. However, there are things to think about enhancing. I've made a TODO list in my initial post of this PR. If you have any suggestions to add up to that list, don't hesitate!

sim590:
Actually, the shutdown process still seems to hang sometimes at random. I'm working on this.

- [ ] Solve shutdown process hanging bug

sim590:
The code has reached a point where it properly does what it wants to do. I'll be working on more tests to integrate to the benchmark script. Notably, I want to check what happens when:
- [ ] multiple nodes starting one after the other and put a value on the network. Are the values still on the network after 10 minutes? How much traffic is triggered by this?
- [ ] replacing clusters successively.

src/dht.cpp
```diff
@@ -286,6 +305,43 @@ Dht::RoutingTable::depth(const RoutingTable::const_iterator& it) const
...
+    auto sortedBucketInsert = [&](const Bucket &b) {
+            for (auto n : b.nodes) {
+                auto here = std::find_if(nodes.begin(), nodes.end(), [&id,&n](std::shared_ptr<Node> &node) {
+                        return id.xorCmp(node->id, n->id) < 0;
+                        });
+                nodes.insert(here, n);
+            }
+        };
+
+    if (bucket != this->end()) {
```
aberaud: already checked on line 313

include/opendht/dht.h
```diff
@@ -824,10 +850,12 @@ class Dht {
     }
 
     void storageAddListener(const InfoHash& id, const InfoHash& node, const sockaddr *from, socklen_t fromlen, uint16_t tid);
-    ValueStorage* storageStore(const InfoHash& id, const std::shared_ptr<Value>& value);
+    ValueStorage* storageStore(const InfoHash& id, const std::shared_ptr<Value>& value, time_point created=time_point::max());
     void expireStorage();
     void storageChanged(Storage& st, ValueStorage&);
 
+    int maintainStorage(InfoHash id, bool force=false, DoneCallback donecb=nullptr);
```
aberaud: return type should be unsigned or size_t (it's a count)

sim590:
I've rebased and squashed all commits into 3 seperate ones:

1. All the new API members are added without any extra functionnality for backward compatibility;
2. Considerable changes were made to the python benchmark script: new test frame and better IPC handling between nodes run as sub process and the main script.
3. The code bringing data persistence into the DHT.

This will let the before vs after test easy to do.

sim590:
Since this has been merged from 4bee627, I'm closing this pull request.

sim590:
Since this has been merged from 4bee627, I'm closing this pull request.

Issue #26, opened 2015-12-21
Master branch doesn't build. GCC incompatibility with clang.


FellowTraveler:
Stack Overflow says:  _"In gcc stl system_clock duration is nanoseconds, but in clang it is microseconds."_

$ make
Scanning dependencies of target opendht
[  4%] Building CXX object CMakeFiles/opendht.dir/src/utils.cpp.o
/Users/au/Projects/opendht/src/utils.cpp:29:49: error: no viable conversion from
      **'time_point<[...], duration<[...], ratio<[...], 1000000000>>>' to 'const**
      **time_point<[...], duration<[...], ratio<[...], 1000000>>>'**
    return std::chrono::system_clock::to_time_t(std::chrono::system_clock::now() + 
                                                ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/au/bin/clang+llvm-3.5.0-macosx-apple-darwin/bin/../include/c++/v1/chrono:750:29: note: 
      candidate constructor (the implicit copy constructor) not viable: no known conversion from
      'time_point<std::__1::chrono::system_clock, typename common_type<duration<long long,
      ratio<1, 1000000> >, duration<long long, ratio<1, 1000000000> > >::type>' to 'const
      std::__1::chrono::time_point<std::__1::chrono::system_clock,
      std::__1::chrono::duration<long long, std::__1::ratio<1, 1000000> > > &' for 1st argument
class _LIBCPP_TYPE_VIS_ONLY time_point
                            ^
/Users/au/bin/clang+llvm-3.5.0-macosx-apple-darwin/bin/../include/c++/v1/chrono:750:29: note: 
      candidate constructor (the implicit move constructor) not viable: no known conversion from
      'time_point<std::__1::chrono::system_clock, typename common_type<duration<long long,
      ratio<1, 1000000> >, duration<long long, ratio<1, 1000000000> > >::type>' to
      'std::__1::chrono::time_point<std::__1::chrono::system_clock,
      std::__1::chrono::duration<long long, std::__1::ratio<1, 1000000> > > &&' for 1st argument
class _LIBCPP_TYPE_VIS_ONLY time_point
                            ^
/Users/au/bin/clang+llvm-3.5.0-macosx-apple-darwin/bin/../include/c++/v1/chrono:772:13: note: 
      candidate template ignored: disabled by 'enable_if' [with _Duration2 =
[...]

PR #27, opened 2015-12-22
Minor cmake fixes

justusranvier:
Ensure the library is built with thread support
Bump Mgspack required version to reflect use of 1.2 features.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,3 +41,7 @@ Makefile.in
 *.l[ao]
 *~
 *.pc
+
+# KDevelop
+.kdev4/
+*.kdev4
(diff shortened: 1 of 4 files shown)


NicoHood:
Does this also apply to the normal makefile? It should be patched there as well, if it wasnt yet.

justusranvier:
I'm more unfamiliar with autotools than with cmake, but I believe the most recent commit does the same thing there.

justusranvier:
The Travis failures do not appear to be related to this commit, but rather with: https://github.com/savoirfairelinux/opendht/issues/26

PR #29, opened 2016-01-06
Refactoring Dht::sendClosestNodes

sim590:
Since 4bee6274, there was some duplication of code around finding the set of closest nodes of an id, I've refactored Dht::sendClosestNodes to use Dht::RoutingTable::findClosestNodes.


sim590:
Closing this as it will be part of the solution for #42.

PR #30, opened 2016-01-07
Search refill optimisation

sim590:
In order to optimize the refill of a search with good `SearchNode`, the refill is now done at each call of `searchStep()`. A refill will try to add more good nodes to the search until the search has reached `SEARCH_NODES` good nodes.

- [x] Include a notion of delay between two `Search::refill` calls to avoid network flood.

@aberaud: Please let know what you think. I'll squash some of the commits before you merge this when it is acceptable.

### DIFF
--- a/src/dht.cpp
+++ b/src/dht.cpp
@@ -860,7 +860,7 @@ Dht::Search::insertNode(std::shared_ptr<Node> node, time_point now, const Blob& 
             removeExpiredNode(now);
 
             auto farthest_good_node = std::find_if(nodes.rbegin(), nodes.rend(),
-                    [=](std::shared_ptr<Node>& n) { return n->isGood(now); }
+                    [=](const SearchNode& n) { return n.node->isGood(now); }
                     );
             if (farthest_good_node != nodes.rend()) {
                 nodes.erase(farthest_good_node.base());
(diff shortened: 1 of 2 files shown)


src/dht.cpp
```diff
@@ -1048,40 +1059,31 @@ Dht::searchStep(Search& sr)
...
         {
-            unsigned added = 0;
-            if (not sr.refilled) {
-                added = sr.refill(sr.af == AF_INET ? buckets : buckets6, now);
-                sr.refilled = true;
+            DHT_ERROR("[search %s IPv%c] expired", sr.id.toString().c_str(), sr.af == AF_INET ? '4' : '6');
+            // no nodes or all expired nodes
+            sr.expired = true;
+            // reset refilled since the search is now expired.
+            sr.refilled = false;
```
aberaud: If the refill succeeds, then refilled will be true. If a second is needed later, it would not happen.

src/dht.cpp
```diff
@@ -939,6 +944,12 @@ Dht::searchStep(Search& sr)
 
     /* Check if the first TARGET_NODES (8) live nodes have replied. */
     if (sr.isSynced(now)) {
+
+        if (sr.nodes.size()-sr.getNumberOfCandidates(now) < SEARCH_NODES and not sr.refilled) {
```
aberaud: Why restricting this to the synced state ? Synced means all 8 first known nodes have answered, so in that case there would be no need to refill at all.

src/dht.cpp
sim590: Indeed. I think we can even get rid of `Search::refilled` member. However, there should be some delay between two refills in order to avoid traffic catastrophies.

src/dht.cpp
sim590: Yes. You are right. I'll fix that.

sim590:
I've included a delay between two successive refills. However, a stronger empirical analysis should be done around the accurate value to use.

@aberaud: What do you think?

sim590:
This change will be included in a PR for #42. Closing this PR.

sim590:
I think this should be merged outside of #52 since it doesn't have to do with refactoring... However, I have to reopen the PR since [I closed and changed rebased in the mean time][]

[I closed and changed rebased in the mean time]: https://github.com/isaacs/github/issues/361

PR #33, opened 2016-01-14
Benchmark python test cleanup

sim590:
Some minor code cleanups. I've also introduced the use of decorators for handling tasks executing before each tests.


sim590:
This also fixes a missing import error from 77c480e.

PR #34, opened 2016-02-15
Change ascii value to letter value in log

kaldoran:
As far as ASCII is concerned, using they decimal value of letter / symbol, should not be a problem.
However, some (linux?) distrubution do not base they char on ascii.

Even thought, ASCII is a standard, and may not evolved, that change will make the program more evolution-proof toward those change.

In fact, that's not a bug, but only an undesired futur. (If ASCII change then log may (try to) print unpritable caracter).

### DIFF
--- a/include/opendht/utils.h
+++ b/include/opendht/utils.h
@@ -90,7 +90,7 @@ struct LogMethod {
     void logPrintable(const uint8_t *buf, size_t buflen) const {
         std::string buf_clean(buflen, '\0');
         for (size_t i=0; i<buflen; i++)
-            buf_clean[i] = buf[i] >= 32 && buf[i] <= 126 ? buf[i] : '.';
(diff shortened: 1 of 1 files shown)


yomgui1:
Hi kaldoran,
What do you think about std::ctype::isgraph() (or more suitable)?

sim590:
@kaldoran Your patch should be based on the `master` branch since it doesn't concern the `index` branch in particular. I'll close this pull request. I suggest you `cherry-pick` your changes onto `master`and open another PR targeted to the `master` branch referencing this old PR.

kaldoran:
isgraph may not be a good option cause space is not considered like a printable caracter.
However std::ctype::isprint() may be a good option.

yomgui1:
Great, I  let you do the change.

PR #36, opened 2016-02-15
Change ascii value to std::isprint() function

kaldoran:
This pull request refere to #34. 

I change the ascii comparaison by the std::ctype::isprint() function, which one is more accurate to the problem.


sim590:
It's minor change and it seems alright. It should assure a better compatibility between all platforms. Merging!

PR #37, opened 2016-02-15
Fix the autotool Makefile, missing pht include

kaldoran:
I fixe the Makefile.am from src, cause we were missing the pht file in it.
Now we can create the makefile using autotools again.

( cmake was already update )

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -12,7 +12,8 @@ libopendht_la_SOURCES = \
         crypto.cpp \
         securedht.cpp \
         dhtrunner.cpp \
-        default_types.cpp
+        default_types.cpp \
+        indexation/pht.cpp
 
 if WIN32
 libopendht_la_SOURCES += rng.cpp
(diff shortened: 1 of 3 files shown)


src/Makefile.am
```diff
@@ -28,4 +29,5 @@ nobase_include_HEADERS = \
         ../include/opendht/securedht.h \
         ../include/opendht/dhtrunner.h \
         ../include/opendht/default_types.h \
-        ../include/opendht/rng.h
+        ../include/opendht/rng.h \
+		../include/opendht/indexation/pht.h
```
sim590: Make sure that lines are aligned together :3

kaldoran: I saw that, tabulation instead of space, i'll edit that

kaldoran:
@sim590 Here i need you for merge if you are ok with it

sim590:
Alright @kaldoran. Since this is a small change, you could squash all of your commits into a single one. Then, I'll merge it.

kaldoran:
@sim590 that's done :)

PR #39, opened 2016-02-18
Prefix bug - A bit may remain in a prefix

kaldoran:
Hi,

There is a little bug in the Prefix code.
If you try to get a prefix that is not a multiple of 8, you will get some extra bits. (In fact you will get X % 8 extra bit), even if they are not print they are here.

Here is an example that illustrate that problem, and confirm it
http://pastebin.com/92q0ie6u

For the simplicity of my test i've change the `auto n=size_ / 8` by `auto n=2`.
Here is the output after recompile all things : 
`./test`
`01010101110000000` 
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;`^ extra bit here`

So here is this pull request.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,3 +45,13 @@ Makefile.in
 # KDevelop
 .kdev4/
 *.kdev4
+
+# Python
+*.pyc
+
+# git backup files
+*.orig
+
+# vim swap files
+*.swp
+*.swo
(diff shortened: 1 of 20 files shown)


kaldoran:
Oops i update the commit ( need remplace ll of them by 1 )

sim590:
You appear to have made your PR to be merged on `master`. You should make a pull request to the `index` branch because we don't want to pull all the `index` branch onto master. Closing this.

kaldoran:
Ooops double sorry then, i fix the list of commit and then i'll create a new one

Issue #45, opened 2016-02-29
[autotools] make uninstall doesn't remove python files


kaldoran:
Hi,

Here is the problem, when i use autotools [ Autogen script and configure script ], then make, and then sudo make install, all goes right.

But if you try to use `make uninstall` (as sudoers ), the command succeed but some files are still here (probably all are still here).

Here is the output of `make uninstall`
http://pastebin.com/a3WBC8dG
At the end you could also see the `ls` that shows that all file are still here.

sim590:
On what revision (commit) have you tried this?

kaldoran:
Last commit of Master branch.

sim590:
Your log seems to indicate you're using a makefile configured on a different branch. It seems you were on branch `index` since the keywords `pht.h` and `pht_cache.h` appear. Especially since the word `pht_cache.h` is there, I can't say if your makefile is alright. Can you make sure to run the following on master branch and see if include files are still not removed.
```sh
$ make clean
$ ./autogen.sh
$ ./configure
$ make
$ sudo make install
$ sudo make uninstall
$ find /usr/include -name \*opendht\*
```
As of the master branch, *autotools doesn't seem to delete installed python files*. This could be improved by adding a rule `uninstall-local:` to clean corresponding python files.

kaldoran:
First, i was trying of another branch, but i re-did it into master and got the same problem.
[ i did not update the pastebin file]. 
I'll update it after this message

kaldoran:
Ok, so between my 2 tests i manually remove the file that was giving trouble.
And now, i've done what you said above and no file are generate.

sim590:
Alright. I'll modify your issue name and leave it opened for the case of python files which are not removed when doing `make uninstall` as I said above.

kaldoran:
Since the PR had been merge i could close this issue :)

Zibri:
make uninstall
make: *** No rule to make target 'uninstall'.  Stop.

sim590:
@Zibri: I suspect, you're using CMake. This thread is about Autotools.

PR #46, opened 2016-02-29
[autotools] pip to uninstall python files

sim590:
As highlighted in #45, autotools files didn't provide a way for uninstalling opendht python module. So I've written a solution for this using `pip`.

@aberaud: since python naming convention differ from platform to platform, I've chosen to check for the binary `pip3` for explicitely specifying the python version and since our python module is written in Python3. Tell me if this way is okay with you.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -16,21 +16,27 @@ AM_PROG_AR
 dnl Check for Doxygen
 AC_ARG_ENABLE([doc], [AS_HELP_STRING([--disable-doc], [Disable documentation generation (doxygen)])])
 AS_IF([test "x$enable_doc" != "xno"], [
-  AC_CHECK_PROGS([DOXYGEN], [doxygen])
-  AS_IF([test -z "$DOXYGEN"],[AC_MSG_WARN([Doxygen not found - continuing without Doxygen support])])
-])
+       AC_CHECK_PROGS([DOXYGEN], [doxygen])
+       AS_IF([test -z "$DOXYGEN"],[AC_MSG_WARN([Doxygen not found - continuing without Doxygen support])])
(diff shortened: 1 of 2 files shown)


sim590:
@aberaud: So as we discuessed, I looked for a way to uninstall a python package using `setup.py` script, [but there's not such thing](http://stackoverflow.com/questions/1550226/python-setup-py-uninstall). However, as highlighted in the previous link, it is possible to produce a *record* file which contains the list of all paths to delete. I think we could fallback on this method if `pip` is not found on the system, but we have to keep in mind that if `pip` is not installed, the following would fail to uninstall:

```sh
$ git clone https://github.com/savoirfairelinux/opendht/ opendht_first
$ cd opendht_first
$ ./autogen && ./configure --enable-python && make && sudo make install

$ cd ..
$ git clone https://github.com/savoirfairelinux/opendht/ opendht_second
$ cd opendht_first
$ ./autogen && ./configure --enable-python
$ sudo make uninstall
```

@aberaud: So do you like keeping `pip` and falling back to the *record* if present?

sim590:
I've added a fallback to the *record*, but after testing it seems a bit less robust (leaving some directories behind and letting `pip` think that the package is still installed) and obviously introduces danger: it will remove any file listed in the `install_record.txt` file. If the file was corrupt in someway and that, for instance, a line would contain an unwanted path, bad things could happen (`/`, `/usr/lib/`)... We simply don't handle this case and it would make more sense to let `pip` handle this.

After all, I think that the record could be generated, but the `make uninstall` command should not try to uninstall based on it. It should let the user do it manually, again only if `pip` is not available. @aberaud, what do you think?

PR #49, opened 2016-03-03
Revert "[Bug] Correction of function commonBits"

sim590:
Reverts savoirfairelinux/opendht#48

After rechecking, this change didn't make sense.


kaldoran:
C.f last comment on the #48 for the purpose of this revert.

PR #55, opened 2016-04-04
Search progressive refill optimisation

sim590:
This is a rebase of an old and closed PR (#30). It was planned to be part of another PR, but on a second thought, it wouldn't be clear enough. So I'm reopenning the PR.

@aberaud: can you review this?

### DIFF
--- a/src/dht.cpp
+++ b/src/dht.cpp
@@ -948,6 +948,12 @@ Dht::searchStep(Search& sr)
 
     /* Check if the first TARGET_NODES (8) live nodes have replied. */
     if (sr.isSynced(now)) {
+
+        if (sr.nodes.size()-sr.getNumberOfCandidates(now) < SEARCH_NODES and not sr.refilled) {
+            auto added = sr.refill(sr.af == AF_INET ? buckets : buckets6, now);
+            DHT_WARN("[search %s IPv%c] refilled with %u nodes", sr.id.toString().c_str(), (sr.af == AF_INET) ? '4' : '6', added);
+        }
+
         if (not sr.callbacks.empty()) {
             // search is synced but some (newer) get operations are not complete
             // Call callbacks when done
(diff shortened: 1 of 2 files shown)


src/dht.cpp
```diff
@@ -859,9 +857,16 @@ Dht::Search::insertNode(std::shared_ptr<Node> node, time_point now, const Blob&
...
-                nodes.pop_back();
+        if (nodes.size()-num_bad_nodes > SEARCH_NODES) {
+            removeExpiredNode(now);
+
+            auto farthest_good_node = std::find_if(nodes.rbegin(), nodes.rend(),
+                [=](const SearchNode& n) { return n.node->isGood(now); }
+            );
+            if (farthest_good_node != nodes.rend()) {
+                nodes.erase(farthest_good_node.base());
+            }
```
aberaud: Note: a node may be good **and** candidate

src/dht.cpp
```diff
@@ -941,8 +946,20 @@ Dht::searchStep(Search& sr)
     DHT_DEBUG("[search %s IPv%c] step", sr.id.toString().c_str(), sr.af == AF_INET ? '4' : '6');
     sr.step_time = now;
 
+    /*
+     * The accurate delay between two refills has not been strongly determined.
+     * TODO: Emprical analysis over refill timeout.
+     */
+    if (sr.refill_time + Node::NODE_EXPIRE_TIME < now and sr.nodes.size()-sr.getNumberOfBadNodes(now) < SEARCH_NODES) {
```
aberaud: We try to insert new nodes in the search every time we hear about them.
What is the case that we try to solve here ?
This will run every 10 minutes, how will it help in that scenario ?

sim590: You can refer to #30 for brief and vague reminder. The following should be more complete:

*The goal was to fasten the "refill" of a search, i.e. the replacement of bad nodes in a wider range around the target hash of the search.*

Before this, the refill of a search was done upon the expiration of a `Search` (all nodes in a search would be expired). We decided to enhance the solution by using the approach where a `Dht::Search::refill` would be done each call of `Dht::Searchstep` always if *last refill* was made at least 10 minutes before. A delay was necessary in case of critical scenario where a considerable amount of nodes would expire all around the target hash of the Search, therefor over refilling the search with potentially wrong nodes (i.e. too far from target hash compared to others). This scenario could be observed in one of our benchmark tests (`benchmark.py --persistence -t delete`, or `-t replace`). This change would make the refill progressive and not wait for a `Search` to expire.

As said in the comment at line 950, the wait time has been arbitrary determined. However, I'm mostly certain there was a reasoning behind choosing 10 minutes, but I can't remember right now. But as you're saying

src/dht.cpp
sim590: True. Fixed it in 12f4139.

PR #56, opened 2016-04-06
Adding cache into indexation

kaldoran:
And here comes the cache for the indexation process.

In this PR i also move the `__normalConditionTest(self)` function of the file  python/tools/dht/tests.py
cause i've planned ahead for a next test which would test all the feature with a real network ( with some node leaving and some getting back etc ).

If that gave trouble i could update my PR

### DIFF
--- a/include/opendht/indexation/pht.h
+++ b/include/opendht/indexation/pht.h
@@ -47,6 +47,16 @@ struct Prefix {
         return Prefix(*this, len);
     }
 
+    /**
+     * Method for getting the state of the bit at the position pos.
+     * @param pos : Pos of the needed bit
+     * @return : true if the bit is at 1
+     *           false otherwise
+     */
+    bool isActivBit(size_t pos) const {
+        return ((this->content_[pos / 8] >> (7 - (pos % 8)) ) & 1) == 1;
+    }
+
     Prefix getFullSize() { return Prefix(*this, content_.size()*8); }
 
     /**
@@ -136,7 +146,7 @@ class Pht {
 
 public:
     /* This is the maximum number of entries per node. This parameter is
-     * critical and influences the traffic alot during a lookup operation.
+     * critical and influences the traffic a lot during a lookup operation.
      */
     static constexpr const size_t MAX_NODE_ENTRY_COUNT {16};
 
@@ -166,6 +176,125 @@ public:
     void insert(Key k, Value v, Dht::DoneCallbackSimple cb = {});
 
 private:
+    class Cache {
+    public:
+        /**
+         * Insert all needed node into the tree according to a prefix
+         * @param p : Prefix that we need to insert
+         */
+        void insert(const Prefix& p) {
+            size_t i = 0;
+            auto now = clock::now();
+
+            std::shared_ptr<Node> curr_node;
+
+            while ( ( leaves_.size() > 0 && leaves_.begin()->first + NODE_EXPIRE_TIME < now )
+                    || leaves_.size() > MAX_ELEMENT ) {
+
+                leaves_.erase(leaves_.begin());
+            }
+
+            if ( !(curr_node = root_.lock()) ) {
+
+                /* Root does not exist, need to create one*/
+                curr_node = std::make_shared<Node>();
+                root_ = curr_node;
+            }
+            
+            curr_node->last_reply = now;
+
+            /* Iterate through all bit of the Blob */
+            for ( i = 0; i < p.size_; i++ ) {
+
+                /* According to the bit define which node is the next one */
+                auto& next = ( p.isActivBit(i) ) ? curr_node->right_child : curr_node->left_child;
+
+                /** 
+                 * If lock, node exists
+                 * else create it
+                 */
+                if (auto n = next.lock()) {
+                    curr_node = std::move(n);
+                } else {
+                    /* Create the next node if doesn't exist*/
+                    auto tmp_curr_node = std::make_shared<Node>();
+                    tmp_curr_node->parent = curr_node;
+                    next = tmp_curr_node;
+                    curr_node = std::move(tmp_curr_node);
+                }
+
+                curr_node->last_reply = now;
+            }
+
+            /* Insert the leaf (curr_node) into the multimap */
+            leaves_.emplace(std::move(now), std::move(curr_node) );
+        }
+
+        /**
+         * Lookup into the tree to return the maximum prefix length in the cache tree
+         *
+         * @param p : Prefix that we are looking for
+         * @return  : The size of the longest prefix known in the cache between 0 and p.size_
+         */
+        int lookup(const Prefix& p) {
+            int pos = 0;
+            auto now = clock::now(), last_node_time = now;
+            
+            /* Before lookup remove the useless one [i.e. too old] */
+            while ( leaves_.size() > 0 &&  leaves_.begin()->first + NODE_EXPIRE_TIME < now ) {
+                leaves_.erase(leaves_.begin());
+            }
+
+            auto next = root_;
+            std::shared_ptr<Node> curr_node;
+
+            while ( auto n = next.lock() ) {
+                /* Safe since pos is equal to 0 until here */
+                if ( (unsigned) pos >= p.size_ ) break;
+
+                curr_node = n;
+                last_node_time = curr_node->last_reply;
+                curr_node->last_reply = now;
+
+                /* Get the Prefix bit by bit, starting from left */
+                next = ( p.isActivBit(pos) ) ? curr_node->right_child : curr_node->left_child;
+
+                ++pos;
+            }
+
+            if ( pos > 0 ) {
+                auto to_erase = leaves_.find(last_node_time);
+                if ( to_erase != leaves_.end() ) 
+                    leaves_.erase( to_erase );
+
+                leaves_.emplace( std::move(now), std::move(curr_node) );
+            } 
+
+            return --pos;
+        }
+
+
+    private:
+        static constexpr const size_t MAX_ELEMENT {1024};
+        static constexpr const std::chrono::minutes NODE_EXPIRE_TIME {5};
+
+        struct Node {
+            time_point last_reply;           /* Made the assocation between leaves and leaves multimap */
+            std::shared_ptr<Node> parent;    /* Share_ptr to the parent, it allow the self destruction of tree */
+            std::weak_ptr<Node> left_child;  /* Left child, for bit equal to 1 */
+            std::weak_ptr<Node> right_child; /* Right child, for bit equal to 0 */
+        };
+
+        std::weak_ptr<Node> root_;                         /* Root of the tree */
+
+        /**
+         * This mutlimap contains all prefix insert in the tree in time order
+         * We could then delete the last one if there is too much node
+         * The tree will self destroy is branch ( thanks to share_ptr )
+         */
+        std::multimap<time_point, std::shared_ptr<Node>> leaves_;
+    };
+
     /**
      * Linearizes the key into a unidimensional key. A pht only takes
      * unidimensional key.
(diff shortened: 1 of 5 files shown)


kaldoran:
@sim590 here i need you for a review :yum:

include/opendht/indexation/pht.h
```diff
@@ -47,6 +47,15 @@ struct Prefix {
         return Prefix(*this, len);
     }
 
+    /**
+     * Method for getting the bit at the position pos.
+     * @param pos : Pos of the needed bit
+     * @return : bit at the position 'pos'
+     */
+    int getBit(size_t pos) const {
```
aberaud: bool

include/opendht/indexation/pht.h
```diff
@@ -166,6 +175,125 @@ class Pht {
     void insert(Key k, Value v, Dht::DoneCallbackSimple cb = {});
 
 private:
+    class Cache {
+    public:
+        /**
+         * Insert all needed node into the tree according to a prefix
+         * %param p : Prefix that we need to insert
+         */
```
sim590: This is a minor observation: the format of documentation should be consistent with the rest of the opendht code, i.e. with [doxygen format][]. I think `%param` is not part of the standard format. Use `@param` instead. 

[doxygen format]: https://www.stack.nl/~dimitri/doxygen/manual/docblocks.html

include/opendht/indexation/pht.h
```diff
@@ -166,6 +175,125 @@ class Pht {
...
+            leaves_.insert(std::pair<time_point, std::shared_ptr<Node>>(now, curr_node ));
+        }
+
+        /**
+         * Lookup into the tree to return the maximum prefix lenght in the cache tree
+         *
+         * @param p : Prefix that we are looking for
+         * @return  : The size of the longest prefix known in the cache between 0 and p.size_
+         */
+        int lookup(const Prefix& p) {
```
sim590: Again, this is minor, but you may review english in the comments.

- line 212 `s/existe/exists/`
- line 218 `s/not exist/doesn't exist/`
- line 228 `s/leave/leaf/`
- line 233 `s/lenght/length/`
- line 281 `s/betwean/between/`
- ...

python/tools/dht/tests.py
```diff
@@ -371,6 +478,9 @@ def _gottaGetThemAllPokeNodes(self, consumer, hashes, nodes=None):
                 for n in DhtFeatureTest.foreignNodes:
                     nodes.add(n)
 
+    def __normalConditionTest(self):
```
sim590: This is odd since `DhtFeatureTest` inherits from `FeatureTest`, so it has `self._normalConditionTest`. Therefor, at line 574, the call is simply:
```python
self._normalConditionTest()
```
as before except that the name of the function changed. In short, `__normalConditionTest` is useless.

src/Makefile.am
```diff
@@ -30,4 +30,4 @@ nobase_include_HEADERS = \
         ../include/opendht/dhtrunner.h \
         ../include/opendht/default_types.h \
         ../include/opendht/rng.h \
-        ../include/opendht/indexation/pht.h
+		../include/opendht/indexation/pht.h
```
sim590: Tab issues. Since I know you use `vim`, you should make sure to `:set noexpandtab` before inserting a tab in an automake file.

src/indexation/pht.cpp
```diff
@@ -9,14 +9,14 @@ const ValueType IndexEntry::TYPE = ValueType::USER_DATA;
...
-        std::shared_ptr<unsigned> max_common_prefix_len, bool all_values)
+        std::shared_ptr<unsigned> max_common_prefix_len, int start, bool all_values)
 {
     struct node_lookup_result {
         bool done {false};
         bool is_pht {false};
     };
 
-    auto mid = (*lo + *hi)/2;
+    auto mid = /*(start >= 0) ? start :*/ (*lo + *hi)/2;
```
sim590: I'm sure this is not meant to be left commented out. ¯\(°_o)/¯

src/indexation/pht.cpp
```diff
@@ -180,10 +188,10 @@ void Pht::insert(Key k, Value v, Dht::DoneCallbackSimple done_cb) {
                 entry.prefix = kp.content_;
                 entry.name = name_;
 
-                updateCanary(*final_prefix);
-                dht_->put(final_prefix->hash(), std::move(entry), done_cb);
+            updateCanary(*final_prefix);
+            dht_->put(final_prefix->hash(), std::move(entry), done_cb);
```
sim590: Bad indentation.

tools/dhtnode.cpp
```diff
@@ -96,7 +96,7 @@ void cmd_loop(std::shared_ptr<DhtRunner>& dht, std::map<std::string, dht::indexa
             break;
         } else if (op == "h" || op == "help") {
             print_help();
-            continue;
+            continue;;
```
sim590: Does this compile? If so, tell me and revert this change anyway.

src/indexation/pht.cpp
```diff
@@ -27,21 +27,26 @@ void Pht::lookupStep(Prefix p, std::shared_ptr<int> lo, std::shared_ptr<int> hi,
         }
         else if (is_leaf or *lo > *hi) {
             // leaf node
+            auto to_insert = p.getPrefix(mid);
+            cache_.insert(to_insert);
+
```
sim590: Beware empty trailing white spaces. See [this][] or [my vimrc](https://github.com/sim590/dotfiles/blob/master/vimrc#L137).

[this]: http://vim.wikia.com/wiki/Remove_unwanted_spaces

include/opendht/indexation/pht.h
kaldoran: Yup i saw that, in fact i use %param on my own doc in C and forget to change it here :)
I'll make an update for that then

src/indexation/pht.cpp
kaldoran: This is not, this is in fact the main part of the code that's commented here. oops :x

tools/dhtnode.cpp
kaldoran: yup that's compile, that's interpreted as a : 
continue;
;
But there is still an error

sim590:
Trying to compile with CMake, I get this error:
```
[ 88%] Linking CXX executable dhtchat
../libopendht.so: undefined reference to `dht::indexation::Pht::Cache::NODE_EXPIRE_TIME'
```
I think [this](http://stackoverflow.com/a/30002401) is related. Also, there are a couple of warnings at compilation time which you could get rid of. For e.g.:
```
/home/simon/prog/opendht/src/indexation/pht.cpp: At global scope:
/home/simon/prog/opendht/src/indexation/pht.cpp:12:62: warning: unused parameter ‘start’ [-Wunused-parameter]
         std::shared_ptr<unsigned> max_common_prefix_len, int start, bool all_values)
                                                              ^
/home/simon/prog/opendht/src/indexation/pht.cpp:151:18: warning: unused parameter ‘ok’ [-Wunused-parameter]
         [=](bool ok){
                  ^
/home/simon/prog/opendht/src/indexation/pht.cpp:175:50: warning: unused parameter ‘values’ [-Wunused-parameter]
         [=](std::vector<std::shared_ptr<Value>>& values, Prefix p) {
```
If the parameters are unused, just remove the name in the lambda/function signature.

include/opendht/indexation/pht.h
```diff
@@ -166,6 +175,125 @@ class Pht {
...
+                root_ = curr_node;
+            }
+            
+            curr_node->last_reply = now;
+
+            /* Iterate through all bit of the Blob */
+            for ( i = 0; i < p.size_; i++ ) {
+
+                /* According to the bit define which node is the next one */
+                auto& next = ( p.getBit(i) ) ? (*curr_node).right_child : (*curr_node).left_child;
```
sim590: You should use the syntax `pointer->element` instead of `(*pointer).element`.
It's simpler and more consistent with the rest of opendht code.

kaldoran: Ok thanks for the review, the patch will come soon

kaldoran: For the previous error, since i compile with autotool etc i'll try it with cmake and correct the error. Thanks.

include/opendht/indexation/pht.h
```diff
@@ -166,6 +176,125 @@ class Pht {
...
+                    tmp_curr_node->parent = curr_node;
+                    next = tmp_curr_node;
+                    curr_node = std::move(tmp_curr_node);
+                }
+
+                curr_node->last_reply = now;
+            }
+
+            /* Insert the leaf (curr_node) into the multimap */
+            leaves_.insert(std::pair<time_point, std::shared_ptr<Node>>(now, curr_node ));
```
sim590: Here's some c++11 tips. You can do this:
```cpp
leaves_.insert(std::pair<time_point, std::shared_ptr<Node>>(now, curr_node));
```
but, you can shrink it to:
```cpp
leaves_.insert(std::make_pair(now, curr_node));
```
furthermore, you can do:
```cpp
leaves_.emplace(now, curr_node);
```
Finally, *if I'm not mistaken*, the leaf would be copied constructed from `curr_node`. Then, you can enhance the code by making `std::multimap::emplace` use the move constructor of `std::shared_ptr`:
```cpp
leaves_.emplace(now, std::move(curr_node));
```

kaldoran:
Here is the lastest commit with all correction said above.

sim590:
@kaldoran: If you can't see the warnings when compiling and you're using autotools, you have to add the flag `--enable-debug` when running `./configure`. Otherwise, using CMake, you should see the warnings. The warnings I've talked you about are not related to the changes you've made except for one that was due to [code commented out by mistake](https://github.com/savoirfairelinux/opendht/blob/c4e090ced73eb79b7c16c5ea853d75f99c37ad0e/src/indexation/pht.cpp#L19). So don't mind removing the warnings unrelated to your PR. I'll get rid of them.

As for the error, yes, even if the build directory is clean, it happens. Make sure you have "tools compilation" enabled since it happens when trying to link with the library and tools do link with the library:

**CMake**
```sh
cmake -DOPENDHT_TOOLS=on
```
**autotools**
Make you run the `configure` script without the option `--disable-tools`.

kaldoran:
@sim590 I see the warning, but i can't reproduce this error : 
> [ 88%] Linking CXX executable dhtchat
> ../libopendht.so: undefined reference to `dht::indexation::Pht::Cache::NODE_EXPIRE_TIME'

Even by adding some flags or whatever.
Even if i try with cmake or autotools.

kaldoran:
The compilation error had been removed thanks to the lastest commit. 
Thanks for your help @sim590

python/tools/dht/tests.py
```diff
@@ -17,6 +17,15 @@
...
 import networkx as nx
+import subprocess
+import re
+
+from matplotlib.ticker import FuncFormatter
+import math
+
+import numpy as np
+import matplotlib.pyplot as plt
+from matplotlib.ticker import FuncFormatter
```
sim590: This diff shows duplicate imports. Probably happenned while resolving conflicts. You should revert this change (lines 20-28).

python/tools/dht/tests.py
sim590: Also, I see that you have copied the code from `_totallyNormalTest` into another function (`_normalConditionTest`) as if you wanted to rename it, but you have not deleted the [old one](https://github.com/savoirfairelinux/opendht/pull/56/files#diff-7590f949505d55358a56a80e319fe3baR597) (line 597).

sim590: @kaldoran, the changes done on `_totallyNormalTest` don't seem to be used, are not complete and don't quite fit into your PR. May be you could revert this, only keep the changes relevant to the PR and make another PR that involves the complete changes around `_totallyNormalTest` that you wanted to make.

python/tools/dht/tests.py
kaldoran: Update done :) Thanks for the review.
And yup in fact that was a trouble will merging ( i suppose )

python/tools/dht/tests.py
kaldoran: Well, i'm going to revert the change then, the update will follow in a new PR

kaldoran:
Here is the final version of my PR, i merge all the change into 2 single commit

sim590:
Merged in 9269bd1.

kaldoran:
Wel :( i didnt understand why you didnt use the merge button ?
Cause in the commit history, there is a double addition of my line ( once by you, once by me) 
https://github.com/savoirfairelinux/opendht/commits/index

sim590:
@kaldoran: I saw some trailing white spaces in many files, so instead of asking you to remove them since I asked a lot yet, I pulled the branch, removed white spaces and merged in manually, hence the commit with double authorship.

PR #65, opened 2016-04-29
dht: prevent bad function call in Value::Filter::chain

sim590:
The following code snippet shows how you can use `Value::Filter::chain` in an unsafe way:
```cpp
#include <opendht.h>
int main(int argc, char *argv[]) {
    auto f = dht::Value::Filter::chain({}, {});
    if (f)
        f({});
    return 0;
}
```
Running this code will cause `bad_function_call` exception. Requiring to check both parameters of `Value::Filter::chain` before calling it is nuisance. Making the check in `chain` methods is better. The new behavior of the `chain` methods are logical in the sense that if a filter is not callable, then it doesn't filter anything, so this is equivalent of returning `true`.

### DIFF
--- a/include/opendht/value.h
+++ b/include/opendht/value.h
@@ -131,8 +131,15 @@ struct Value
         using std::function<bool(const Value&)>::function;
     public:
         static Filter chain(Filter&& f1, Filter&& f2) {
-            return [f1,f2](const Value& v){
-                return (f1 ? f1(v) : true) && (f2 ? f2(v) : true);
+            if (not f1 and f2)
+                return f2;
+            else if (f1 and not f2)
+                return f1;
+            else if (not f1 and not f2)
+                return {};
+            else
+                return [f1,f2](const Value& v){
+                    return f1(v) && f2(v);
             };
         }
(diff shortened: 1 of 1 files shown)


include/opendht/value.h
```diff
@@ -132,7 +132,7 @@ struct Value
     public:
         static Filter chain(Filter&& f1, Filter&& f2) {
             return [f1,f2](const Value& v){
-                return f1(v) && f2(v);
+                return (f1 ? f1(v) : true) && (f2 ? f2(v) : true);
```
aberaud: for better performance this could be simplified: chain could just return f1 or f2 if the other is not set (same for bellow).

sim590:
```cpp
if (not f1 and f2)
    return f2;
else if (f1 and not f2)
    return f1;
else if (not f1 and not f2)
    return {};
else
    return [f1,f2](const Value& v){
        return f1(v) && f2(v);
    };
```
Do you mean something like this ?

sim590:
Closing this, since the issue is solved in 1af71e8.

PR #67, opened 2016-05-10
crypto: mark all operator bool as explicit

yomgui1:
If operator bool is not marked as explicit and you try to do such:

a == b

where a and b are instances of class with overloaded bool operator,
the behavior is not the waited one: a and b are casted to bool
BEFORE the comparaison! This could lead into non waited situation.

This patch fixes all operator bool() where overloaded.
This also implements PublicKey operators == and != as needed in value.h.
This last showed a typical case explained upper.

Signed-off-by: Guillaume Roguez <guillaume.roguez@savoirfairelinux.com>

### DIFF
--- a/include/opendht/crypto.h
+++ b/include/opendht/crypto.h
@@ -67,7 +67,7 @@ struct PublicKey
     PublicKey(PublicKey&& o) noexcept : pk(o.pk) { o.pk = nullptr; };
 
     ~PublicKey();
-    operator bool() const { return pk; }
+    explicit operator bool() const { return pk; }
 
     PublicKey& operator=(PublicKey&& o) noexcept;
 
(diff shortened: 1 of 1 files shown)


include/opendht/crypto.h
```diff
@@ -94,8 +94,14 @@ struct PublicKey
     PublicKey(const PublicKey&) = delete;
     PublicKey& operator=(const PublicKey&) = delete;
     void encryptBloc(const uint8_t* src, size_t src_size, uint8_t* dst, size_t dst_size) const;
+
+	friend bool operator==(const PublicKey& lhs, const PublicKey& rhs);
```
sim590: @yomgui1: There are some prefix spaces you could remove. Also, why don't you declare-implement those two functions in one place?

aberaud:
Merged in c5da946e76f6df53be97f5bd5e205d5f2afd95af

include/opendht/crypto.h
yomgui1: about the non-member declaration:

https://herbsutter.com/2013/05/20/gotw-4-class-mechanics/

see item 5. It's an encapsulation guideline

Issue #43, opened 2016-02-18
DHT queries


sim590:
Traffic usage has not been concern as of yet, but with data persistence functionality, it is a major issue that has to be addressed. In order to do so, here is the idea of filtering values by field values and field themselves. The following table represents a `Dht::Value`:

| Id  | ValueType | Data | OwnerPk | RecipientHash | UserType | Signature | EncryptedData |
| --- | --------- | ---- | ------- | ------------- | -------- | --------- | ------------- |
| ... | ...       | ...  | ...     | ...           | ...      | ...       | ...           |
| ... | ...       | ...  | ...     | ...           | ...      | ...       | ...           |
| ... | ...       | ...  | ...     | ...           | ...      | ...       | ...           |

Operations `Dht::get` and `Dht::put` could optionally ask for a parameter specifying filters over rows and columns of the above table. *However, this feature depends entirely on the completion of #42, since two separate queries over a same hash would have to be distinguished since they would not be necessarely satisfied in the same way.*

sim590:
I have been working on this. [Here](https://github.com/sim590/opendht/commits/queries)'s what I have done so far. Packing, unpacking and remote handling of queries is done. What's missing is the part where queries are actually locally processed and sent to nodes. Handling of the response, i.e. associating the query's response to the callback and filter to apply, is also missing. The following is summarizing an abstract model of how to handle this.

![conception](https://github.com/sim590/opendht/raw/conception/doc/conception/queries/concept.png)

**NOTE**
- A `Query` distinguishes one `Get` structure instance from another. The same is true for `LocalListener` structure;
- Upon receiving an answer for a query, you only call callbacks for which the associated query is satisfied by the query to which the response is targeted;
- Request structure doesn't actually have a field `std::shared_ptr q`. The query shall only be captured into the `Request::done_cb` lambda.
- In order to track which `SearchNode` has already been sent a given query, we keep a pointer of query associated to a network engine request using a map:

  ```cpp
  std::map<std::shared_ptr<Query>, std::shared_ptr<Request>>
  ```
  That's why I use `std::shared_ptr<Query>`.

@aberaud: Do you have any thoughts? In the meantime, I'll be implementing this.

sim590:
The code has reached a mature state. I'll be opening a PR shortly.

sim590:
[...]

PR #70, opened 2016-05-19
[DHT] Queries: remote values filtering

sim590:
This is addressing #43. The main functionnality is operationnal. I'm still working on the following:

- [x] Add documentation in the wiki about queries (how to use, use cases, benefits, ...);
- [x] Handle spaces and SELECT,WHERE lower/upper case in constructing a query from a string: `Query::Query(std::string&&)`;

### DIFF
--- a/include/opendht/value.h
+++ b/include/opendht/value.h
@@ -456,15 +456,11 @@ struct FilterDescription
     };
 
     FilterDescription() {}
-    FilterDescription(Field t) : field(t) {}
+    FilterDescription(Field f) : field(f) {}
 
     Field getField() { return field; }
 
-    template <typename Packer>
-    void msgpack_pack(Packer& p) {
-        p.pack_map(1);
-        p.pack(std::string("t")); p.pack(field);
-    }
+    bool operator==(FilterDescription& fd) { return field == fd.field; }
 
     void msgpack_unpack(msgpack::object msg) {
         if (auto t = findMapValue(msg, "t"))
@@ -481,7 +477,17 @@ protected:
  * This is meant to narrow data to a set of specified fields. This is equivalent
  * to a SQL "SELECT" statement.
  */
-using FieldFilterDescription = FilterDescription;
+struct FieldFilterDescription : public FilterDescription
+{
+    FieldFilterDescription() {}
+    FieldFilterDescription(Field f) : FilterDescription(f) {}
+
+    template <typename Packer>
+    void msgpack_pack(Packer& p) {
+        p.pack_map(1);
+        p.pack(std::string("t")); p.pack(field);
+    }
+};
 
 /*!
  * @class   ValueFilterDescription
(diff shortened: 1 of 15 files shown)


sim590:
- [ ] We have to be clarify this:  https://github.com/sim590/opendht/commit/f46043069d60040081bf18668d4714b19c3e2316#commitcomment-17565515

@aberaud: I think this is ready for review. When you're done, I'll squash everything in relevant seperate commits and you'll be clear to merge.

sim590:
Travis is complaining, but it doesn't appear to be related to my changes. Or else I don't understand the error message...

tools/dhtscanner.cpp
```diff
@@ -63,7 +63,7 @@ step(DhtRunner& dht, std::atomic_uint& done, std::shared_ptr<NodeSet> all_nodes,
         done--;
         std::cout << done.load() << " operations left, " << all_nodes->size() << " nodes found." << std::endl;
         cv.notify_one();
-    });
+    }, {}, {});
```
sim590: @aberaud: can you comment on this? It makes Travis happy, but the [signature](https://github.com/sim590/opendht/blob/queries/include/opendht/dhtrunner.h#L65) it's using should allow empty parameter... I don't really like this as it more looks like a work around the issue.

include/opendht/value.h
```diff
@@ -422,6 +481,255 @@ struct Value
...
+     *  [SELECT <$field$> [WHERE <$field$=$value$>]]
+     *
+     *  where
+     *
+     *  - $field$ = *|id|value_type|owner_pk|user_type
+     *  - $value$ = $string$|$integer$
+     *  - $string$: a simple string WITHOUT SPACES.
+     *  - $integer$: a simple integer.
+     */
+    Query(std::string&& q_str);
```
aberaud: const std::string& q_str (value is not moved in method)

sim590:
A result of the choice of implementation is that when creating a selection in a query, like so:
```cpp
Query q("SELECT id")
dht.get(hash, [](std::vector<std::shared_ptr<Value>>>& values) {
        /* only the field id is valid and relevant */
    }, {}, {}, std::move(q));
```
Some fields in the returned values may or may not be empty. In reality, if a value is fetched from the local node's storage, the value will have other non-empty valid fields and if another value is fetched from a remote node's storage, only the selected fields specified in the query will be valid. **This is something that has to be documented, or else a change in the API is needed if we want to provide unambiguous API**. By that I mean provide a different callback signature that would offer a consistent structure containing the results of the selection of a query...

Also, Let's say you make a `get` request with the query `SELECT id` for the target hash `h` using the `dhtnode` tool and that it happens the local node has the storage for `h`. When stacking values for the `get` callback before ef067e6, the following would happen:
```
Get: found value (after 0.000437637s)
	Value[id:7f96f114f92b4ce9 Data (type: 0 ): 66 6f 6f ]
Get: found value (after 0.00424051s)
	Value[id:7f96f114f92b4ce9 Data (type: 0 ): ]
Get: completed (took 0.00561377s)
```
The first value is from the local storage and the second is from the network (data is empty as a result of the query `SELECT id`). If OpenDHT allows empty data and that two different values from different peers can have the same value id, there is no way to differentiate the first value from the second one.

ef067e6 makes sure that no redundant information will be present, i.e. that in the example above only the first value would be present in the callback `values` argument.

sim590:
Merged. See 51aa1de.

PR #73, opened 2016-06-01
dht: consider IPv4 or IPv6 disconnected on operation done

sim590:
Put and Get wrapped done callbacks need to consider the case only ipv4 **or** ipv6 connection is available. In that case, don't block waiting for the disconnected network and call done callbacks if the operation succeeded on one of the networks.

This bug was introduced by both d44c590 ('put' donecb wouldn't be called) and bd08506 ('get' donecb wouldn't be called)

### DIFF
--- a/src/dht.cpp
+++ b/src/dht.cpp
@@ -731,8 +731,6 @@ void
 Dht::searchStep(std::shared_ptr<Search> sr)
 {
     if (not sr or sr->expired or sr->done) return;
-    if (sr->nodes.empty() and (sr->af == AF_INET ? buckets : buckets6).isEmpty())
-        return; // wait for connection
 
     const auto& now = scheduler.time();
     DHT_LOG.DEBUG("[search %s IPv%c] step (%d requests)", sr->id.toString().c_str(), sr->af == AF_INET ? '4' : '6', sr->currentGetRequests());
--- a/src/dht.cpp
+++ b/src/dht.cpp
@@ -822,8 +822,10 @@ Dht::searchStep(std::shared_ptr<Search> sr)
                     a.callback(true, sr->getNodes());
                     a.callback = nullptr;
                 }
-                ait = sr->announce.erase(ait);
-                continue;
+                if (not a.permanent) {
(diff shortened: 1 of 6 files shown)


sim590:
[1] Let *table* be either the ipv4 or ipv6 table such that *table* is empty. `Announce` and `Get` structures won't be cleared until new nodes are added to the *table*. Also, done callbacks associated to the *table* won't even be called with failure status.

I think this is not a desired behavior. Callbacks should called and the associated structures cleared from the search. This PR was written this way in order to exactly not conflict with d44c590 and bd08506.

@aberaud: But, d44c590 and bd08506 exactly do imply [1], doesn't it? I'd appreciate you comment on that so that I can update my fix if need be.

aberaud:
All searches should end and the done callback called asap, if there is no node they should fail immediately, so we should find another way to postpone operations performed before the first node is contacted or have answered.

sim590:
Taking another look at this and after our discussion, the changes in d44c590 and bd08506 that introduced a bug were to solve a race condition where `SecureDht` would put its certificate on the network while routing table bootstraping was not complete yet. So, in order to solve all of this in a clean way, the option to make a *permanent* put is introduced (back). This should solve the race condition problem. The changes in d44c590 and bd08506 that introduced the bug were reverted.

@aberaud: any comment?

src/dht.cpp
```diff
@@ -1411,7 +1412,10 @@ Dht::put(const InfoHash& id, std::shared_ptr<Value> val, DoneCallback callback,
     auto done6 = std::make_shared<bool>(false);
     auto donecb = [=](const std::vector<std::shared_ptr<Node>>& nodes) {
         // Callback as soon as the value is announced on one of the available networks
-        if (callback && !*done && (*done4 && *done6)) {
+        auto done46 = (*done4 and buckets6.isEmpty()) or
+                      (*done6 and buckets.isEmpty()) or
+                      (*done4 and *done6);
+        if (callback && !*done && done46) {
```
aberaud: not related

sim590:
Merged. See 54a4327.

PR #77, opened 2016-06-14
pht: fix invalid comparison, inexact match lookup

sim590:
This is a change around the code doing the _inexact match lookup_. The algorithm should check the number of common bits between the target prefix and candidate prefix found in the leaves. However, by error, the code was comparing the target prefix with the wrong data: the indexed hash rather than the prefix key for that index entry.

@kaldoran, @aberaud: It's been a while since I worked on this part of the code. Can you review this?


kaldoran:
Forget to comment about it yesterday but all seams ok for me.

Issue #61, opened 2016-04-19
[PHT] complete nodes splitting


sim590:
Let `n_0` be a leaf in the PHT and `I_0`, a node on the DHT. When `n_0` is *full*, the next index insert involving `n_0` and performed by `I_0` will *split* the node in two leaves. As of now, the split consists only of `I_0` creating two *children* beneath `n_0`, i.e. two new leaves. `I_0` will then create a new entry in one of both new leaves and end the insert operation. However, left in this state, the "split" is not done yet. In theory, all entries in `n_0` should be "moved" to the corresponding new leaves. This is a mandatory behavior of the theoretical PHT.

### Solution

A simple and elegant way, suggested by @aberaud, would be to make use of the `listen` feature available from the DHT node. The idea is the following:

Let `p` a PHT key and `p_0` the longest prefix of `p` existing on the PHT. After inserting an index entry at `p_0`, i.e. `put(HASH(p_0), index_entry)`, you `listen` to the children `p_1`, `p_2` of `p_0` so that if `p_1` or `p_2` becomes a leaf, the index entry on `p_0` will be inserted to the associated child.

However, one should consider that:

- An index entry should not last longer on the PHT after being "moved" than it would have without, i.e. when "moving" the entry, a time stamp should be transferred. The signature of `put` let you do that by using the `created` parameter;
- *Moving an entry*: an entry is never moved, but only copied. Entries expire themselves.

sim590:
@aberaud, do you have any thoughts to add?

sim590:
@kaldoran: I think that such an enhancement make the PHT cache from #41 better.

kaldoran:
Should have update this one a lot sooner but here comes the update.
Lot of version of this thing was done, with some try and error, a great version will came out son.

We finally choose to split until there is a difference between they key, and split until the point of splitting instead of split it recursively that will generate a lot of trafic.

For the merge, since the current version of merge is not relevant i'm going to change the method.
Instead of counting value in a triangle shape ( parent, current node and is sibling ) we are just going to stop update a canary when a put is done, only a split will update it.

kaldoran:
So here is some update, @aberaud we speak about the idea of not updating the canary when we put another value.
But if you put some value with the same start, then the tries going to grow until the end of the common bits into thoses prefix.
Then, after some time, the canary will expire, but some value are still in the node where the split append, but for now they are unreachable ...

I try to use another method here, where i update the canary above the node where it split but not seams to work ..
[...]

PR #78, opened 2016-06-15
[PHT] Key consistency

sim590:
This PR enhances the `Pht` API:

- A `Key` is now `std::map<std::string, Blob>` rather than `std::map<std::string, Prefix>`.
- The `Pht` now depends on a key specification

        using KeySpec = std::map<std::string, size_t>

  This key specification sets the order, the number and the length of the fields in the key.

Other minor changes:

- `dhtnode` will handle the key spec by setting the keyspec for the pht with the first key entered by the user;
- the benchmark has been updated accordingly.

### DIFF
--- a/include/opendht/indexation/pht.h
+++ b/include/opendht/indexation/pht.h
@@ -65,12 +65,7 @@ struct Prefix {
      * @return The prefix of this sibling.
      */
     Prefix getSibling() const {
-        Prefix copy = *this;
-        if (size_) {
-            size_t last_bit =  (8 - size_) % 8;
-            copy.content_.back() ^= (1 << last_bit);
-        }
-        return copy;
+        return swapBit(size_ - 1);
     }
 
     InfoHash hash() const {
@@ -113,6 +108,27 @@ struct Prefix {
         }
 
         return 8 * i + j;
+    }
+
+    /**
+     * This method swap the bit a the position 'bit' and return the new prefix
+     *
+     * @param bit Position of the bit to swap
+     *
+     * @return The prefix with the bit at position 'bit' swapped
+     *
+     * @throw out_of_range Throw out of range if bit does not exist
+     */
+    Prefix swapBit(size_t bit) const {
+        if ( bit >= size_ )
+            throw std::out_of_range("bit larger than prefix size.");
+
+        Prefix copy = *this;
+
+        size_t offset_bit = (8 - bit) % 8;
+        copy.content_[bit / 8] ^= (1 << offset_bit);
+
+        return copy;
     }
 
     size_t size_ {0};
@@ -174,6 +190,7 @@ public:
     {
         lookup(k, [=](std::vector<std::shared_ptr<Value>>& values, Prefix) { cb(values); }, done_cb, exact_match);
     }
+
     /**
      * Adds an entry into the index.
      */
@@ -236,7 +253,16 @@ private:
      */
     virtual Prefix linearize(Key k) const {
         if (not validKey(k)) { throw std::invalid_argument(INVALID_KEY); }
-        return Blob {k.begin()->second.begin(), k.begin()->second.begin() + keySpec_.begin()->second};
+
+        Prefix p = Blob {k.begin()->second.begin(), k.begin()->second.end()};
+
+        auto bit_loc = p.size_ + 1;
+        for ( auto i = p.content_.size(); i <= keySpec_.begin()->second; i++ ) {
+            p.content_.push_back(0);
+            p.size_ += 8;
+        }
+
+        return p.swapBit(bit_loc);
     };
 
     /**
--- a/include/opendht/indexation/pht.h
+++ b/include/opendht/indexation/pht.h
@@ -53,7 +53,7 @@ struct Prefix {
      * @return : true if the bit is at 1
      *           false otherwise
      */
-    bool isActivBit(size_t pos) const {
+    bool isActiveBit(size_t pos) const {
         return ((this->content_[pos / 8] >> (7 - (pos % 8)) ) & 1) == 1;
     }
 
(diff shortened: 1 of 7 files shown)


kaldoran:
Strange that travis can't compile it, got no problem on local copy of your rep.
going to review change into next comment

sim590:
@kaldoran: Travis doesn't use the same compiler version as yours then. And now, travis is happy. 

@kaldoran, @aberaud : Since this is an API change, I would like some feedback. What do you think about the key spec implementation? I also have moved `Pht::linearize` into `public:`. I did this to provide a way to generate a prefix from a `Key` for displaying in the dhtnode. Do you think `Pht::linearize` should stay `private:`? Actually, more I think about it and more I think that `Prefix` should not be exposed in any way and only keys should be exposed in the API since a  `Prefix` is literally the Pht internal representation of a `Key`. The user shouldn't be aware of the existence of the `Prefix` class. What do you think?

sim590:
I've been thinking about hiding `Prefix` from the API and I think this is not a good thing since it would prevent the user to cumulate information about the state of the PHT. For e.g., the `Prefix` makes it possible to draw the [trie][] formed on network.

- [ ] However, there should be a more user friendly signature that doesn't show the `Prefix`:

    ```cpp
    using LookupCallbackSimple = std::function<void(std::vector<std::shared_ptr<Value>>& values)>;
    using LookupCallback = std::function<void(std::vector<std::shared_ptr<Value>>& values, Prefix p)>;

    void lookup(Key k, LookupCallback cb = {}, DoneCallbackSimple doneCb = {}, bool exact_match = true);
    void lookup(Key k, LookupCallbackSimple cb = {}, DoneCallbackSimple doneCb = {}, bool exact_match = true) {
        lookup(k, [=](std::vector<std::shared_ptr<Value>>& values, Prefix p) { cb(values); }, doneCb, exact_match);
    }
    ```

[trie]: https://en.wikipedia.org/wiki/Trie

include/opendht/indexation/pht.h
```diff
@@ -296,26 +218,38 @@ class Pht {
...
      *
-     * @return return The linearized key.
+     * @return the prefix of the linearized key.
      */
-    static Prefix linearize(Key k) {
-        if (k.size() != 1) { throw std::invalid_argument("PHT only supports unidimensional data."); }
-        return k.begin()->second;
+    virtual Prefix linearize(Key k) const {
+        if (not validKey(k)) { throw std::invalid_argument(INVALID_KEY); }
+        return Blob {k.begin()->second.begin(), k.begin()->second.begin() + keySpec_.begin()->second};
```
kaldoran: Going to get an error if the key go a size < to the keySpec lenght ;)

tools/dhtnode.cpp
```diff
@@ -277,15 +293,20 @@ void cmd_loop(std::shared_ptr<DhtRunner>& dht, std::map<std::string, dht::indexa
...
-            indexes.at(index).insert(createPhtKey(parseStringMap(keystr)), v,
-                [=](bool success) {
-                    if (not success) {
-                        std::cout << "Pht::insert: failed." << std::endl;
-                        return;
+            try {
+                auto key = createPhtKey(parseStringMap(keystr));
+                indexes.at(index).insert(key, v,
+                    [=](bool success) {
+                        if (not success) {
```
kaldoran: Should use same method as at line 280
Pht::insert: (success ) ? "done" : "failed"

include/opendht/indexation/pht.h
kaldoran: Edit : After trying we do not get an error, however .. we cannot retrive the real Prefix since we get the prefix followed by 0.
We should use that we spoke about : a 1 and then fill with 0.

I will update this PR to add those modification since a need them for the issue #61

kaldoran:
Since my PR #85 was closed ( and totaly need to be close ) 
There is the change i've made, you can cherry-pick here 
https://github.com/kaldoran/opendht/commits/pht-api-fix
Commit : 9e9bdd0d8a638b62ab31ff805de8eeb4f6df48ab

kaldoran:
Since you correct a mispell into the function isActiveBit, and since the bug is the same as was into PR #85 , could you also cherry-pick the commit 
62f77bb1e6a130eda09c4f8be4071254c7ebc229

include/opendht/indexation/pht.h
```diff
@@ -115,6 +110,27 @@ struct Prefix {
...
+     * This method swap the bit a the position 'bit' and return the new prefix
+     *
+     * @param bit Position of the bit to swap
+     *
+     * @return The prefix with the bit at position 'bit' swapped
+     *
+     * @throw out_of_range Throw out of range if bit does not exist
+     */
+    Prefix swapBit(size_t bit) const {
+	if ( bit >= copy.size_ )
```
kaldoran: Just saw that right now, there is a tabulation here and line just under this one
Line 123 :  \tif( bit >= copy.size_)
Line 124 :  \t    throw std::out_of_range("bit larger than prefix size.");

Could you please edit that ? 
Or should i do it and then give you the commit to cherry pick ?

kaldoran:
By the look of it, it seams that i gave you the wrong commit number ... 
Sorry for that here is the correct one ( that compile ) 
https://github.com/kaldoran/opendht/commits/pht-api-fix
c00c2073822bad8b0bdbd96bed975f22fbd1c8d1

Not sure if you can remove the previous commit of not.

kaldoran:
I've made all the needed for the compilation error and correct the algorithm of linearize :)
The commit to cherry-pick is updated

kaldoran:
Wow, what append on this compilation but clang succeed and gcc does not, seams that this is a bug due to travis

kaldoran:
I merge it since that compile on clang ( and should do the same of gcc without the pull error on travis ).

PR #84, opened 2016-06-27
[dhtnode] user experience

sim590:
I have fixed a `dhtnode`'s optional argument parsing problem and added a manpage.

### DIFF
--- a/tools/dhtnode.cpp
+++ b/tools/dhtnode.cpp
@@ -29,7 +29,7 @@ extern "C" {
 using namespace dht;
 
 void print_usage() {
-    std::cout << "Usage: dhtnode [-p local_port] [-b bootstrap_host:port]" << std::endl << std::endl;
+    std::cout << "Usage: dhtnode [-v [logfile]] [-i] [-d] [-n network_id] [-p local_port] [-b [bootstrap_host[:port]]]" << std::endl << std::endl;
     std::cout << "dhtnode, a simple OpenDHT command line node runner." << std::endl;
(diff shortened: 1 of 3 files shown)


tools/dhtnode.cpp
```diff
@@ -29,7 +29,7 @@ extern "C" {
 using namespace dht;
 
 void print_usage() {
-    std::cout << "Usage: dhtnode [-p local_port] [-b bootstrap_host:port]" << std::endl << std::endl;
+    std::cout << "Usage: dhtnode [-h] [-v [logfile]] [-i] [-d] [-n network_id] [-p local_port[:port]] [-b [bootstrap_host]]" << std::endl << std::endl;
```
aberaud: - ":port" is not part of "-p"
- "-h" : it's definitely not part of normal usage, can't be used with other options, and I think shouldn't be included.

sim590: Yes. The former is a mistake. 

The latter is a choice, really. Usage doesn't imply "normal usage" IMO, but is meant to show *all* available usage syntaxes. If you run `make --help` for instance, you'll see `Usage: make [options] [target] ...` with the listed options where `-h` appears. But I don't mind. I can remove it.

aberaud: The syntax suggests "-h" can be used along with other options, which is not the case (printing the help message is a special case that makes other flags meaningless).

If it was to be listed (up to your choice), it should have it's own "Usage:" line, like what you did for the manpage.

sim590: I see your point. I'll do it this way for a more clear format.

PR #85, opened 2016-06-27
[PHT] Key consistency with fix

kaldoran:
This is a PR that fix my review on the PR #78 from Sim590.

Thiis PR add a padding on key that are shorter than the keyspec. 
The padding is basically a '1' followed by '0' in binary.

### DIFF
--- a/include/opendht/indexation/pht.h
+++ b/include/opendht/indexation/pht.h
@@ -65,12 +65,7 @@ struct Prefix {
      * @return The prefix of this sibling.
      */
     Prefix getSibling() const {
-        Prefix copy = *this;
-        if (size_) {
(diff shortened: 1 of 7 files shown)


sim590:
Better not fork the discussion and duplicate #78 if the PR subject is the same after all. Just ask me to cherry-pick your changes.

kaldoran:
Well  i was asking @aberaud last week and said that i should just open a new PR fixing that issue.
Btw, the keyspec is the one given by the user, but i have an extra 8 bit for the padding.

Since then, all padding would have the same structure. 
We cannot achieve that without add this padding section.

kaldoran:
Ok no problem ;)

Issue #87, opened 2016-07-01
OpenDHT does not build on armel.


aviau:
OpenDHT does not build on armel, see the following log:
 - https://buildd.debian.org/status/fetch.php?pkg=ring&arch=armel&ver=20160630.2.b3d131f~dfsg1-2&stamp=1467369556

This is the only Debian-supported architecture that OpenDHT does not build on.

Note: I am able to sponsor access to Debian porter machines. See the [documentation](https://dsa.debian.org/doc/guest-account/) on requesting access.

sim590:
I've begun testing on armel archetecture in order to reproduce and find the bug. I'll update shortly.

sim590:
So the issue is definitely [this](https://lists.debian.org/debian-arm/2013/12/msg00007.html), i.e. that `std::future` templates are not supported on armel archetecture.

@aberaud: A solution would be to use the following to disable `std::future` signatures when not supported.
```cpp
#if defined(_GLIBCXX_HAS_GTHREADS) && defined(_GLIBCXX_USE_C99_STDINT_TR1) && (ATOMIC_INT_LOCK_FREE > 1)
```
I know that [ring](https://github.com/savoirfairelinux/ring-daemon/blob/master/src/ringdht/ringaccount.cpp#L241) uses this feature. I'm looking for a solution for this too.

sim590:
@aviau: Actually, if you look at `DhtRunner`, you can see that std::future is used for blocking on the dhtrunner internal ops queue. So, if I were to use the fix above, I'd have to implement handling of `listen` token return in another way which would mean a bit of code to write and it wouldn't be really beautiful with many `ifdef` statements. We have to ask ourselves if this is worth supporting armel archetecture.

aviau:
If the fix isn't too complicated, I would love to see it merged. Otherwise, I could even maintain it in Debian. All of the other Ring dependencies already build on armel. Future armel-incompatible changes will be spotted easily because the Debian CI will constantly build OpenDHT.

aberaud:
Do we **need** to support ARMv5 to be in Debian ?
This architecture is obsolete, every chip is at least ARMv7. If it doesn't support std::future it should really not our problem, we won't re-implement the weel just for that. _OpenDHT requires C++11 support._

Ring also makes use of std::future.

sim590:
> Do we need to support ARMv5 to be in Debian ?
[...]

PR #89, opened 2016-07-02
dhtrunner: no std::future on armel architecture

sim590:
This PR is addressing #87.

As explained on this [debian mailing list](https://lists.debian.org/debian-arm/2013/12/msg00007.html), old ARM architecture cannot use std::future. This patch creates few alternative function definitions for`DhtRunner::listen` functions preserving behavior when with `std::future` support.

*Note: Code compiles perfectly on either architecture. However, I haven't checked if the code runs well. I'll update shortly. Feel free to highlight any error you I could have missed.*

### DIFF
--- a/include/opendht/dhtrunner.h
+++ b/include/opendht/dhtrunner.h
@@ -96,6 +96,7 @@ public:
         getFilterSet<T>());
     }
 
+#if defined(_GLIBCXX_HAS_GTHREADS) && defined(_GLIBCXX_USE_C99_STDINT_TR1) && (ATOMIC_INT_LOCK_FREE > 1)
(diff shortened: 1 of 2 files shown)


aberaud:
This breaks the API and creates a different API between architectures. RIng makes use of this API - and uses std::future, so we would also need to patch Ring for an architecture we (currently) don't support.

sim590:
Yes, you're right. This solution is really not ideal.

aberaud:
This doesn't provide the same behavior and would only work in threaded mode, so for instance it wouldn't work in Ring (where OpenDHT is in non-threaded mode), because it would deadlock waiting for the token.

OpenDHT will aim to support architectures where C++11 is available but won't reimplement the standard library for others. Closing this PR.

PR #90, opened 2016-07-02
doc: Fix install path of man page

zorun:
Man pages should go in /usr/share/man, not /usr/man:

  http://www.pathname.com/fhs/pub/fhs-2.3.html#USRSHAREMANMANUALPAGES


sim590:
Thanks for the patch. I did do this to quickly.

sim590:
Thanks for the patch. I did do this to quickly.

PR #93, opened 2016-07-19
dht: return consistent query from local storage

sim590:
When querying from local storage with SELECT f, a FieldValueIndex with empty
fields was returned without considering the SELECT statement requiring only a
field "f".


kaldoran:
Woops, meant to merge not close :D

PR #94, opened 2016-07-22
dht: make main store a vector<unique_ptr<Storage>>

sim590:
Since the elements inside are going to be moved constantly during the life of the node, this design is better. Also, it works around a crash when the store is resized (elements inside moved) at some point (always happenned the 3rd time) for some reason. This bug was known when compiled with LLVM >7.3.0 (>clang-703.0.31).

### DIFF
--- a/include/opendht/dht.h
+++ b/include/opendht/dht.h
@@ -406,7 +406,7 @@ private:
     RoutingTable buckets {};
     RoutingTable buckets6 {};
 
-    std::vector<Storage> store;
+    std::vector<std::unique_ptr<Storage>> store;
     size_t total_values {0};
     size_t total_store_size {0};
     size_t max_store_size {DEFAULT_STORAGE_LIMIT};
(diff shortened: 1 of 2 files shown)


sim590:
I'm linking this issue to Ring's tuleap bug tracker's issue about this: https://tuleap.ring.cx/plugins/tracker/?aid=556

kaldoran:
Since Travis does not compile i've tried on my own computer. This need some fix
`dht.cpp: In member function ‘dht::NetworkEngine::RequestAnswer dht::Dht::onGetValues(std::shared_ptr<dht::Node>, dht::InfoHash&, dht::want_t, const dht::Query&)’:
dht.cpp:2728:15: error: no match for ‘operator!=’ (operand types are ‘std::unique_ptr<dht::Dht::Storage>’ and ‘std::vector<std::unique_ptr<dht::Dht::Storage> >::iterator {aka __gnu_cxx::__normal_iterator<std::unique_ptr<dht::Dht::Storage>*, std::vector<std::unique_ptr<dht::Dht::Storage> > >}’)
     if ((*st) != store.end() && not (*st)->empty()) {
`

sim590:
Yes. I did resolve conflicts too quickly. Now it should be fine.

kaldoran:
Seams ok now that compile on my computer. I'll merge

Issue #75, opened 2016-06-10
Different libraries built with Autotools and CMake


szotsaki:
When building with Autotools, the following library files are created:
- libopendht.so.0.0.0 [executable, valid shared library with correct so naming]
- @libopendht.so.0 [link to the former one with one major version]
- @libopendht.so [link to the first one, only useful for development purposes]

These are correct, and the package library also gets its name from the .so: _libopendht0_.

When bulding with CMake, the following is created:
- libopendht.so [executable, with incorrect so naming]
- @symlinks are missing

aberaud:
Thanks for reporting,
Should be fixed by https://github.com/savoirfairelinux/opendht/commit/9615774114abc8b59eb37ef5b68eccbaf3862d34

szotsaki:
This works great, thank you! 

Please, be aware that the Autotools version generates .so with version `0.0.0` while CMake with `0.6.1`.

szotsaki:
Just to be sure: should I open a new ticket regarding the Autotools version difference? Thank you!

sim590:
@szotsaki. You're right, Autotools doesn't have correct version naming. I've reopened this issue.

szotsaki:
With the release of 0.6.2 I rebuilt the package but unfortunately CMake still builds `/usr/lib64/libopendht.so` without any suffix and symlinks.

sim590:
In fact, it wasn't part of the 0.6.2 release. This should be part of a future release where this whole issue is fixed (including autotools). We did release 0.6.2 for covering the critical #73 bug.

kaldoran:
Here is an update :

>  Start with version information of ‘0:0:0’ for each libtool library.
Update the version information only immediately before a public release of your software. 
[...]
> Never try to set the interface numbers so that they correspond to the release number of your package. 

Source : https://www.gnu.org/software/libtool/manual/html_node/Updating-version-info.html#Updating-version-info

Then a fix could be adding `-release @PACKAGE_VERSION@` line 7 of src/Makefile.am.
Which then going to create libopendht-0.6.X.so ( x = 2 for now ).

Not sure if this is the correct way to deal with that problem.

sim590:
I have looked a little bit more into this. The main issue is that CMake and Libtool have different and conflicting shared object versionning system. If we want consistent version releases, we're kind of forced to do like [this](https://github.com/chewing/libchewing/issues/55).

carlwgeorge:
[...]

PR #95, opened 2016-07-25
autotools: versionning consistent with CMake

sim590:
Addressing #75, this patch brings consistent versionning between Autotools and CMake.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1,9 +1,20 @@
-AC_INIT(opendht, 0.6.1)
+dnl define macros
+m4_define([opendht_major_version], 0)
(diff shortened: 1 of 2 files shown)


carlwgeorge:
It looks like this fix was undone by https://github.com/savoirfairelinux/opendht/commit/73df3082507d936f1acf3b0a1b6ae328aec3cd92#diff-480477e89f9b6ddafb30c4383dcdd705.  That commit dropped the `-version-info` flag, and now autotools creates `libopendht.so.0.0.0` instead of `libopendht.so.1.9.0`.

Issue #91, opened 2016-07-04
DhtRunner does not allow to give the time_point


kaldoran:
While the Dht allow us to give the time_point that we need to put a value into the Dht, the DhtRunner does not allow that.

Shouldn't be a way to put the data using the DhtRunner and add a time_point value ?

kaldoran:
@sim590 and @aberaud i need you there and link that to issue #61 where i need such things

sim590:
So you're asking if the `DhtRunner` API should have function signatures for setting a time of creation for a value to put on the DHT. I'm not sure there'd be a good use case for this and that the `DhtRunner` should provide this. This is more a kind of a low level API. You would have to directly use the `Dht` or `SecureDht` class, but since the `Pht` class model uses a `DhtRunner` instance, then may be the logical thing to do would be to make `Pht` a `friend` of `DhtRunner` so that you could do:

kaldoran:
Thanks for the answer Simon, thats why i was looking for. 
But there is no signature into the SecureDht where you can use a time_point ( or at least i didnt saw it ).
I'm going to follow this : 
`dht_->dht_->put(/* SecureDht::put signature parameters */ );`
Thanks.

I think we can now close this issue.

sim590:
@kaldoran: `SecureDht` is a child class of `Dht`, so it inherits from the `Dht` functions you are looking for.

kaldoran:
@sim590 ok, thanks i will look into that :D 
I will do that and then make a break on the complet node splitting since there is some issue with the merge ( the split look good for the moment )

aberaud:
Hi,

If PHT actually needs this feature, other OpenDHT users might also need it, so it may be relevant to expose the feature in public APIs. PHT is built on top on the DHT so this approach would be cleaner.

Using the private dht_ object directly is not possible as it would create thread safety issues.

sim590:
> Using the private dht_ object directly is not possible as it would create thread safety issues.

Yes! Oops ;)

> If PHT actually needs this feature, other OpenDHT users might also need it, so it may be relevant to expose the feature in public APIs. PHT is built on top on the DHT so this approach would be cleaner.

Yes. I was actually hesitating while giving advice. Considering the thread safety issue, I think this is clear that the `friend` design is not an option.

kaldoran:
I forget to reopen this one :D now its done

PR #97, opened 2016-07-27
Add timepoint into the API of SecureDht and DhtRunner

kaldoran:
Since i need this feature (see #91) for the issue #61, here is the modification that correct #91.

### DIFF
--- a/include/opendht/dhtrunner.h
+++ b/include/opendht/dhtrunner.h
@@ -161,16 +161,16 @@ public:
     void cancelListen(InfoHash h, size_t token);
     void cancelListen(InfoHash h, std::shared_future<size_t> token);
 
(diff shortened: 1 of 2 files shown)


sim590:
Since you're exposing this feature to the main API, you might as well test it and check if it works properly. Actually, this code was written during the work on data persistence. Since it has been a while and we have changed the code alot, you should check if it runs well. May be create a test case with `dhtnode` and point us to a commit number on your branch we could checkout to run the test.

kaldoran:
Well thoses test were a good idea.
Since i've did them and found that we can put outdated data into the Dht :D 
Anyway next PR will correct that

sim590:
I confirm it works. Merged.

Issue #96, opened 2016-07-26
[Dht] Optimize search


kaldoran:
Optimize search by using NodeCache .
That will speed up all search and should optimize network usage.

kaldoran:
Update here : 
I should update `Dht::bootstrapSearch` of dht.cpp [line : 1238] to include cache.

**At the end on modification** : 
Dht::bootstrapSearch -> network_engine [cache] -> NodeCache [Choose iv4 or v6]-> NodeMap -> function which choose the node (or list of node)

Sound good for you @sim590  or @aberaud ?

sim590:
What you said on IRC. ¯\_(ツ)_/¯

PR #101, opened 2016-08-02
[PHT] Optimize search

kaldoran:
Since I open issue #96 I also give this patch to merge.
Now the bootstrap is not done by using the cache instead of the closest bucket (and if needed bucket next to this closest bucket ).

### DIFF
--- a/include/opendht/network_engine.h
+++ b/include/opendht/network_engine.h
@@ -362,10 +362,14 @@ public:
 
     void blacklistNode(const std::shared_ptr<Node>& n);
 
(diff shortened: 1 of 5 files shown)


include/opendht/network_engine.h
```diff
@@ -362,10 +362,14 @@ class NetworkEngine final {
 
     void blacklistNode(const std::shared_ptr<Node>& n);
 
+    std::vector<std::shared_ptr<Node>> getNodesFromCache(const InfoHash& id, sa_family_t sa_f);
```
sim590: Since this method is only a wrapper for `NodeCache::getCachedNodes`, you may as well call thsi method ^ the same.

Issue #71, opened 2016-05-28
[DHT] value pagination


sim590:
A logical enhancement after #43 is to introduce pagination of the values. This will let reduce packet size of responses to `get` requests. The following diagram shows the sequences of a `get` operation:

![sequence diagram](https://github.com/sim590/opendht/raw/conception/doc/conception/pagination/seq-diagram.png)

where
- the SQL-ish statements are describing a query.
- `<q.fields>` and `<q.filters>` respectively denote the fields list and the filters specified by the query the user passed through the API.
- the `id_n` are the ids recovered by the first query. `n` depends obviously on the response to the first request.

@aberaud: what do you think?

sim590:
I've started to work on this [here](https://github.com/sim590/opendht/commits/value-pagination).

PR #107, opened 2016-08-06
dht: more accurate name for requested nodes count

sim590:
The number of currently requested nodes, formerly named currentGetRequests, was renamed since it was not accurately describing the behavior. The number we are counting is not the number of 'get' requests, but rather the number of nodes being requested simultaneously.

### DIFF
--- a/include/opendht/dht.h
+++ b/include/opendht/dht.h
@@ -292,8 +292,8 @@ private:
        the target 8 turn out to be dead. */
     static constexpr unsigned SEARCH_NODES {14};
 
-    /* Concurrent requests during a search */
-    static constexpr unsigned SEARCH_REQUESTS {4};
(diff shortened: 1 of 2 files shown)


include/opendht/dht.h
```diff
@@ -292,8 +292,8 @@ class Dht {
        the target 8 turn out to be dead. */
     static constexpr unsigned SEARCH_NODES {14};
 
-    /* Concurrent requests during a search */
-    static constexpr unsigned SEARCH_REQUESTS {4};
+    /* Concurrent search nodes requested count */
```
aberaud: "Concurrent requests during a search" describes the intended behavior.

sim590: It is equivalent for now, but when the pagination behavior (#71) will be merged, the latter will be more explicit.

PR #113, opened 2016-08-15
dht: value pagination fix

sim590:
This fixes some issues 97b573c did introduce. For instance, some nodes would not expire properly and some of the timings computation were not right.

### DIFF
--- a/src/dht.cpp
+++ b/src/dht.cpp
@@ -201,6 +201,16 @@ struct Dht::SearchNode {
     /**
      * Could a particular "get" request be sent to this node now ?
      *
+     * A 'get' request can be sent when all of the following requirements are
+     * met:
+     *
+     *  - The node is not expired;
+     *  - If we have heard from the node, we must have heard from him in the last
+     *    NODE_EXPIRE_TIME minutes;
+     *  - The request must not already have been sent;
+     *  - No other request satisfying the request must be pending;
+     *  - The pagination process for this particular 'get' must not have begun;
+     *
      * @param now     The time reference to now.
      * @param update  Time of the last "get" op for the search.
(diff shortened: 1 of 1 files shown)


src/dht.cpp
```diff
@@ -209,37 +219,37 @@ struct Dht::SearchNode {
...
-                }
+        const auto& sq_status = std::find_if(getStatus.cbegin(), getStatus.cend(),
+            [&q](const SyncStatus::value_type& s) {
+                return s.first and q and q->isSatisfiedBy(*s.first);
+            }
         );
         return not node->isExpired() and (now > last_get_reply + Node::NODE_EXPIRE_TIME or update > last_get_reply)
-               and ((get_status == getStatus.cend() or not get_status->second)
-                   and (sq_status == getStatus.cend() or not sq_status->second or not sq_status->second->pending()));
+            and not pending(getStatus)
```
sim590: Actually, this line could block some requests that could be sent concurrently. The solution is to use `sq_status`. I did have some trouble with it, so I removed it, but I see it's the logical thing to do.

src/dht.cpp
```diff
@@ -201,45 +201,54 @@ struct Dht::SearchNode {
...
      * @param now     The time reference to now.
      * @param update  Time of the last "get" op for the search.
      * @param q       The query defining the "get" operation we're referring to.
      *
      * @return true if we can send get, else false.
      */
     bool canGet(time_point now, time_point update, std::shared_ptr<Query> q = {}) const {
-        /* find request status for the given query */
-        const auto& get_status = q ? getStatus.find(q) : getStatus.cend();
+        /* FInd request status for the given query */
```
aberaud: Fi

Issue #44, opened 2016-02-19
[PHT] Multi dimensional key support


sim590:
As of now, only unidimensional keys are supported, i.e. keys containing only **one field**. Having support for multidimensionnal keys could enlarge the range of use cases of the PHT indexation feature. This is also a sort of optimization for incomplete queries which could be resolved faster using a [z-curve][] linearization (interleaving bits on diffrent fields of the key) for e.g.

[z-curve]: https://en.wikipedia.org/wiki/Z-order_curve

kaldoran:
In the next few week i will start working on it, with Adrien we start figure out what method would be the better one and a solution comes out. 
I'm going to create a new structure that will encode unknown bit and known one. 
Then the lookupstep going to look for both version a prefix if a value is not given.

kaldoran:
,

kaldoran:
Work in progress on that, for the moment i can transform Prefix on 1 bit to Prefix on 2 bit ( 1 bit indicate if the value is known, the second one indicate the value ) and vice-versa.

Next step : Change the lookupStep and make it work with prefix on 2 bit. ( without changing the algorithm), then change algorithm for lookup on prefix with 2 bit ( with some bit unknown).

PR #115, opened 2016-08-16
Pht: First step for multidimensional key

kaldoran:
This PR is the first step of the multidimensional key support for issue #44.
The second step is to include multidimensional with some unknown value or bit into the multidimensional key to insert on the PHT.

For example on the DhtNode : 
il Profile name:,surname:rey* could look up for a profil with unknown name and a surname that start with 'rey'. [This feature is ready to implement in this code, then next step is to change Pht::lookupStep.

### DIFF
--- a/include/opendht/indexation/pht.h
+++ b/include/opendht/indexation/pht.h
@@ -35,8 +35,9 @@ struct Prefix {
         size_(std::min(first, p.content_.size()*8)),
         content_(Blob(p.content_.begin(), p.content_.begin()+size_/8))
     {
+
         auto rem = size_ % 8;
-        if ( not flags_.empty() ) {
+        if ( not p.flags_.empty() ) {
             flags_ = Blob(p.flags_.begin(), p.flags_.begin()+size_/8);
             if (rem)
                 flags_.push_back(p.flags_[size_/8] & (0xFF << (8 - rem)));
@@ -51,6 +52,7 @@ struct Prefix {
             throw std::out_of_range("len larger than prefix size.");
         if (len < 0)
             len += size_;
+
         return Prefix(*this, len);
     }
 
@@ -77,9 +79,11 @@ struct Prefix {
      * @return The prefix of this sibling.
      */
     Prefix getSibling() const {
-        if ( not size_ )
-            return Prefix(*this);
-        return swapBit(size_);
+        Prefix copy = *this;
+        if ( size_ )
+            copy.swapContentBit(size_ - 1);
+
+        return copy;
     }
 
     InfoHash hash() const {
@@ -137,6 +141,10 @@ struct Prefix {
         return 8 * i + j;
     }
 
+    void swapContentBit(size_t bit) {
+        swapBit(content_, bit);
+    }
+
     /**
      * This method swap the bit a the position 'bit'
      *
@@ -144,11 +152,6 @@ struct Prefix {
      * @return The prefix with the bit at position 'bit' swapped
      * @throw out_of_range Throw out of range if bit does not exist
      */
-    Prefix swapBit(size_t bit) const {
-        if ( bit >= content_.size() * 8 )
-            throw std::out_of_range("bit larger than prefix size.");
-    }
-
     void swapFlagBit(size_t bit) {
         swapBit(flags_, bit);
     }
@@ -183,9 +186,8 @@ private:
 
     std::string blobToString(const Blob &bl) const {
         std::stringstream ss;
-
-        auto bn = size_ % 8;
-        auto n = size_ / 8;
+        auto bn = bl.size() % 8;
+        auto n = bl.size() / 8;
 
         for (size_t i = 0; i < bl.size(); i++)
             ss << std::bitset<8>(bl[i]) << " ";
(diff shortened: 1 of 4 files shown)


include/opendht/indexation/pht.h
```diff
@@ -28,37 +28,61 @@ namespace indexation {
...
+            if (rem)
+                flags_.push_back(p.flags_[size_/8] & (0xFF << (8 - rem)));
+        }
+
         if (rem)
             content_.push_back(p.content_[size_/8] & (0xFF << (8 - rem)));
     }
 
+    /**
+     * Get a prefix of the Prefix
```
sim590: I'd rather call that a "sub prefix"

include/opendht/indexation/pht.h
```diff
@@ -28,37 +28,61 @@ namespace indexation {
...
         return Prefix(*this, len);
     }
 
     /**
-     * Method for getting the state of the bit at the position pos.
-     * @param pos : Pos of the needed bit
-     * @return : true if the bit is at 1
-     *           false otherwise
-     * @throw out_of_range Throw out of range if the bit at 'pos' does not exist
+     * Flags are considered as activ if flag is empty or if the flag
```
sim590: 's,activ,active,g'

include/opendht/indexation/pht.h
```diff
@@ -28,37 +28,61 @@ namespace indexation {
...
+     * at pos 'pos' is activ
```
sim590: same as above.

include/opendht/indexation/pht.h
```diff
@@ -28,37 +28,61 @@ namespace indexation {
...
+     *
+     * @see isActivBit in private function
```
sim590: again.

Signed-off by Grammar Nazi

include/opendht/indexation/pht.h
```diff
@@ -28,37 +28,61 @@ namespace indexation {
...
-    bool isActiveBit(size_t pos) const {
-        if ( pos >= content_.size() * 8 )
-            throw std::out_of_range("Can't detect active bit at pos, pos larger than prefix size or empty prefix");
+    bool isFlagActive(size_t pos) const {
+        return flags_.empty() or isActiveBit(flags_, pos);
+    }
 
-        return ((this->content_[pos / 8] >> (7 - (pos % 8)) ) & 1) == 1;
+    /**
+     * @see isActivBit in private function
```
sim590: Hmmm. You might have to use a search and replace tool. There are lots of them.

include/opendht/indexation/pht.h
```diff
@@ -124,30 +156,128 @@ struct Prefix {
...
+        auto n = bl.size() / 8;
+
+        for (size_t i = 0; i < bl.size(); i++)
+            ss << std::bitset<8>(bl[i]) << " ";
+        if (bn)
+            for (unsigned b=0; b < bn; b++)
+                ss << (char)((bl[n] & (1 << (7 - b))) ? '1':'0');
+
+        return ss.str();
+    }
```
sim590: This method doesn't sound right in Prefix... May be you can make it a lambda inside Prefix::toString since that's the only place you're calling it from.

sim590: Or just put it inside cpp file as static function along with Prefix::toString body definition.

src/indexation/pht.cpp
```diff
@@ -278,35 +288,94 @@ void Pht::insert(Prefix kp, IndexEntry entry, std::shared_ptr<int> lo, std::shar
...
 }
 
+Prefix Pht::zcurve(const std::vector<Prefix>& all_prefix) const {
+    Prefix p;
+
+    if ( all_prefix.size() == 1 )
+        return all_prefix[0];
+
+    /* Loop on all uint8_t of the input prefix */
+    for ( size_t j = 0, bit = 0; j < all_prefix[0].content_.size(); j++) {
```
sim590: Use a variable to describe the value `all_prefix[0].content_.size()`. It will be easier to read. For instance:
```cpp
/* all prefix objects have the same size */
size_t prefix_size = all_prefix[0].content_.size()
```

src/indexation/pht.cpp
```diff
@@ -278,35 +288,94 @@ void Pht::insert(Prefix kp, IndexEntry entry, std::shared_ptr<int> lo, std::shar
...
+                auto f = all_prefix[bit].flags_[j] & mask;
+
+                /* Move this bit at the right position according to the diff
+                   and merge it into content and flags in the same way */
+                content |= ( diff >= 0 ) ? c >> diff : c << std::abs(diff);
+                flags   |= ( diff >= 0 ) ? f >> diff : f << std::abs(diff);
+
+                /* If we are on the last prefix of the vector get back to the first and
+                ,move the mask in order to get the n + 1nth bit */
+                if ( bit == all_prefix.size() - 1 ) { bit = -1; ++i; mask >>= 1; }
```
sim590: Since `bit` is `size_t` it seemed odd to me that you assign it `-1` even if it's ["well defined" by the standard](http://stackoverflow.com/a/28247783). While this code is totally right, I would suggest to use `bit` more consistently with its type by replacing the line above by:
```cpp
if ( bit++ == all_prefix.size() - 1 ) { bit = 0; ++i; mask >>= 1; }
```
And, obviously, remove the `bit` incrementation in the loop header.

PR #116, opened 2016-08-17
pht : Filter of duplicate values

kaldoran:
I add a filter that filter duplicated value : 

Here is a pastebin of what append with the modification 
http://pastebin.com/ZGL1MPew

First i insert 11 same value, then the lookup and there is only one value found (but 11 in storage)
Then i re-add 10 values (we are now above the maximum value in a node that should create a splitting).
Line 56 show that it does not since all values are the same (it also show only one of the 21 values in the storage table)
Finally : 
Line 68 to 72 show that there is 42 values in the storage (21 Canary + 21 values). 

This will be the new behaviour.

### DIFF
--- a/src/indexation/pht.cpp
+++ b/src/indexation/pht.cpp
@@ -28,13 +28,13 @@ static std::string blobToString(const Blob &bl) {
 std::string Prefix::toString() const {
     std::stringstream ss;
 
-    ss << "Prefix : " << std::endl << "\tContent_ : ";
+    ss << "Prefix : " << std::endl << "\tContent_ : \"";
     ss << blobToString(content_);
-    ss << std::endl;
+    ss << "\"" << std::endl;
 
-    ss << "\tFlags_ :   ";
+    ss << "\tFlags_   : \"";
     ss << blobToString(flags_);
-    ss << std::endl;
+    ss << "\"" << std::endl;
 
     return ss.str();
 }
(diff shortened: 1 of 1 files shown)


src/indexation/pht.cpp
```diff
@@ -193,6 +193,14 @@ void Pht::lookupStep(Prefix p, std::shared_ptr<int> lo, std::shared_ptr<int> hi,
                 IndexEntry entry;
                 entry.unpackValue(*value);
 
+                auto it = std::find_if(vals->cbegin(), vals->cend(), [&](const std::shared_ptr<IndexEntry>& ie) {
+                    return ie->value == entry.value;
+                });
```
sim590: Let's say, someone inexes a same value 1024 times. Then, the DHT node won't accept anymore values and at the PHT level, the nodes will refuse to enter the split process, therefor blocking the whole PHT, may be even from the root. I think you should not avoid splitting, but only filter values for the user, i.e. wrapping the `LookupCallback` passed by the user and filter the values there. The prefered place for doing so would be in the `Pht::lookup` function. 

@kaldoran, @aberaud

sim590: This is actually related to a wider problem which deserves to be indexed by an issue. There's a solution that applies on DHT layer.

src/indexation/pht.cpp
```diff
@@ -28,13 +28,13 @@ static std::string blobToString(const Blob &bl) {
 std::string Prefix::toString() const {
     std::stringstream ss;
 
-    ss << "Prefix : " << std::endl << "\tContent_ : ";
+    ss << "Prefix : " << std::endl << "\tContent_ : \"";
     ss << blobToString(content_);
-    ss << std::endl;
+    ss << "\"" << std::endl;
```
aberaud: '"'

src/indexation/pht.cpp
```diff
@@ -28,13 +28,13 @@ static std::string blobToString(const Blob &bl) {
...
 
-    ss << "\tFlags_ :   ";
+    ss << "\tFlags_   : \"";
     ss << blobToString(flags_);
-    ss << std::endl;
+    ss << "\"" << std::endl;
```
aberaud: '"'

PR #52, opened 2016-03-11
[Dht] Network engine

sim590:
This is a major refactoring of the DHT as suggested in #42. Up to now, the `dhtnode` seems to work properly. What's still yet to be done:

### Issues

- [x] fix requests take longer than before;

### Tests
- [x] test performance;
- [x] test network traffic;
- [x] test data persistence behavior;

PR #122, opened 2016-09-01
dht: data persistence 2.0

sim590:
This is a rework of #7 using queries (#43) and the network engine atomic requests (#52). Now, the data persistence behavior will be to first probe for a value ID before sending it, therefor reducing the overall traffic substantially.

Here you can see a graph with the following properties:

- 1024 nodes
- 17.5 put operations per sec (1Ko)
- 8.5 node shutdowns per sec
- 8.5 node joins per sec
- *y axis*: Traffic of the whole network in Kb/s
- *x axis*: Time in seconds

<div style="text-align:center">

</div>

### DIFF
--- a/include/opendht/dht.h
+++ b/include/opendht/dht.h
@@ -512,11 +512,30 @@ private:
     void confirmNodes();
     void expire();
 
+    /**
+     * Generic function to execute when a 'get' request has completed.
+     *
+     * @param status  The request passed by the network engine.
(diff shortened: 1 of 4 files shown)


sim590:
I have to refactor a bit before merge.
- [ ] refactor code

sim590:
Doing some more test, I can see I have to bring a minor change to my code.
- [ ] fix callbacks not called properly.

PR #128, opened 2016-09-13
[dht] concurrent queries

sim590:
This PR is completing some work unintentionaly left undone after the queries. This has left place to some bugs like infinite loops when one or more concurent queries would be issued before a first request would complete.

**This PR introduces the following behavior**:

When multiple queries are queued for a same hash, the search will find the closest node to which send the first queued query. This process will repeat for each queries until all queries are sent to a sufficient number of nodes.

*Note that no more than `MAX_REQUESTED_SEARCH_NODES` nodes will be solicited. However, if 100 requests are queued for a same hash, there could be up to 100x`MAX_REQUESTED_SEARCH_NODES` requests sent at the same time since we only count the solicited nodes and not the requests. This will most of the time not be true because queries are susceptible to satisfy others, hence substantially reducing the number of requests sent. @aberaud: if you think this should change, let me know before merging.*

### DIFF
--- a/src/dht.cpp
+++ b/src/dht.cpp
@@ -265,31 +265,31 @@ struct Dht::SearchNode {
      * met:
      *
      *  - The node is not expired;
-     *  - If we have heard from the node, we must have heard from him in the last
-     *    NODE_EXPIRE_TIME minutes;
-     *  - The request must not already have been sent;
-     *  - No other request satisfying the request must be pending;
      *  - The pagination process for this particular 'get' must not have begun;
+     *  - There hasn't been any response for a request, satisfying the initial
+     *    request, anytime following the initial request.
+     *  - No other request satisfying the request must be pending;
      *
      * @param now     The time reference to now.
-     * @param update  Time of the last "get" op for the search.
+     * @param update  The time of the last 'get' op satisfying this request.
      * @param q       The query defining the "get" operation we're referring to.
      *
      * @return true if we can send get, else false.
      */
     bool canGet(time_point now, time_point update, std::shared_ptr<Query> q = {}) const {
-        /* Find request status for the given query */
-        const auto& get_status = getStatus.find(q);
-        /* Find request status for a query satisfying the initial query */
-        const auto& sq_status = std::find_if(getStatus.cbegin(), getStatus.cend(),
-            [&q](const SyncStatus::value_type& s) {
-                return s.first and q and q->isSatisfiedBy(*s.first) and s.second and s.second->pending();
+        if (node->isExpired() or not (now > last_get_reply + Node::NODE_EXPIRE_TIME or update > last_get_reply))
+            return false;
+
+        auto completed_sq_status {false}, pending_sq_status {false};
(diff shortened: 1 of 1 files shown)


sim590:
I have not yet tested this with the latest merged commit on master. Let me rebase and check if everything works appropriately.

sim590:
I have just checked and I have not notcied any difference in the code's behavior. 

+1

src/dht.cpp
```diff
@@ -265,31 +265,34 @@ struct Dht::SearchNode {
...
      * @param q       The query defining the "get" operation we're referring to.
      *
      * @return true if we can send get, else false.
      */
     bool canGet(time_point now, time_point update, std::shared_ptr<Query> q = {}) const {
         /* Find request status for the given query */
-        const auto& get_status = getStatus.find(q);
+        const auto& completed_sq_status = std::find_if(getStatus.cbegin(), getStatus.cend(),
+            [&q,&update](const SyncStatus::value_type& s) {
+                return s.first and q and q->isSatisfiedBy(*s.first) and s.second and s.second->reply_time > update;
```
aberaud: isSatifiedBy is called twice for every getStatus. Maybe there is a way to call it only once ?

aberaud: (could be performed in a single loop ?)

src/dht.cpp
```diff
@@ -1366,11 +1429,11 @@ unsigned Dht::Search::getNumberOfBadNodes() const {
 }
 
 time_point
-Dht::Search::getLastGetTime() const
+Dht::Search::getLastGetTime(std::shared_ptr<Query> q) const
 {
     time_point last = time_point::min();
     for (const auto& g : callbacks)
-        last = std::max(last, g.second.start);
+        last = std::max(last, (q and q->isSatisfiedBy(*g.second.query) ? g.second.start : time_point::min()));
```
aberaud: What if q is not set

sim590: Indeed!

src/dht.cpp
```diff
@@ -1029,31 +1076,36 @@ Dht::searchSendGetValues(std::shared_ptr<Search> sr, SearchNode* pn, bool update
...
-        if (pn) {
-            if (not pn->canGet(now, up, query))
-                return nullptr;
-            n = pn;
-        } else {
-            for (auto& sn : sr->nodes) {
-                if (sn.canGet(now, up, query)) {
-                    n = &sn;
-                    break;
+        auto get_available_node =
```
aberaud: flatten the lambda

src/dht.cpp
```diff
@@ -1086,10 +1144,13 @@ Dht::searchSendGetValues(std::shared_ptr<Search> sr, SearchNode* pn, bool update
                 paginate(ws, query, n);
         }
 
-        if (not sr->isSynced(now) or cb == sr->callbacks.end())
-            break; /* only trying to find nodes, only send the oldest query */
+        /* We only try to send one request. return. */
+        return n;
```
aberaud: could send all requests for this node ?

PR #130, opened 2016-09-15
introduce SockAddr to store IP addresses

aberaud:

PR #131, opened 2016-09-15
add per IP rate limiting

aberaud:


sim590:
I'll review this after #130.

sim590:
Since #130 has been merged, you could rebase this branch to include only the relevant commit in your PR.

include/opendht/network_engine.h
```diff
@@ -462,8 +464,8 @@ class NetworkEngine final {
     const Logger& DHT_LOG;
 
     NodeCache cache {};
-    std::queue<time_point> rate_limit_time {};
-    static std::mt19937 rd_device;
+    Limiter<MAX_REQUESTS_PER_SEC> rate_limiter {};
+    std::map<SockAddr, Limiter<MAX_REQUESTS_PER_SEC/8>> address_rate_limiter {};
```
sim590: Why divide by 8? I get that 1600 requests per ip may be too high since this is presently set to be a global max, but why just divide by 8? Is it related to `TARGET_NODES` constant? May be you could comment on it.

src/network_engine.cpp
```diff
@@ -191,18 +191,14 @@ NetworkEngine::sendRequest(std::shared_ptr<Request>& request)
...
-    while (not rate_limit_time.empty() and duration_cast<seconds>(now - rate_limit_time.front()) > seconds(1))
-        rate_limit_time.pop();
-
-    if (rate_limit_time.size() >= MAX_REQUESTS_PER_SEC)
+    if (not rate_limiter.limit(now))
         return false;
 
-    rate_limit_time.emplace(now);
-    return true;
+    auto it = address_rate_limiter.emplace(addr, Limiter<MAX_REQUESTS_PER_SEC/8>{});
```
sim590: I see that the dividing operation is done here again. May be there could be a constant `MAX_REQUESTS_PER_IP_PER_SEC` defined in the header?

sim590: What if the entry for the key `addr` is already present in the map? Should it be:
```cpp
auto it = address_rate_limiter.find(addr);
if (it == address_rate_limiter.end()) {
    it = address_rate_limiter.emplace(addr, Limiter<MAX_REQUESTS_PER_SEC/8>{});
}
```

aberaud: emplace will only emplace if the element is not in the map, otherwise will just return the <iterator, bool>. The bool is true if insertion took place.

aberaud: done through documented type definition

include/opendht/network_engine.h
aberaud: done through documented type definition

PR #133, opened 2016-09-21
[dht] bug fixes, timing computation enhancements

sim590:
This is one more consistency fix around operation timings. This PR also fixes a sneaky bug that has appeared in the queries code.

### DIFF
--- a/include/opendht/node.h
+++ b/include/opendht/node.h
@@ -50,6 +50,18 @@ struct Node {
     std::string getAddrStr() const {
         return addr.toString();
     }
+
+    /**
+     * Makes notice about an additionnal authentication error with this node. Up
+     * to MAX_AUTH_ERRORS errors are accepted in order to let the node recover.
+     * Upon this limit, the node expires.
+     */
+    void authError() {
+        if (++auth_errors > MAX_AUTH_ERRORS)
+            setExpired();
+    }
+    void authSuccess() { auth_errors = 0; }
+
     bool isExpired() const { return expired_; }
     bool isGood(time_point now) const;
     bool isPendingMessage() const;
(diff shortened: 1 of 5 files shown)


include/opendht/node.h
```diff
@@ -50,6 +50,16 @@ struct Node {
...
         return addr.toString();
     }
+
+    /**
+     * Makes notice about an additionnal authentication error with this node. Up
+     * to MAX_AUTH_ERRORS errors are accepted in order to let the node recover.
+     * Upon this limit, the node expires.
+     */
+    void authError() {
+        if (++auth_errors > MAX_AUTH_ERRORS)
```
aberaud: auth_errors count is never reset on success

src/network_engine.cpp
```diff
@@ -814,12 +814,14 @@ NetworkEngine::sendListen(std::shared_ptr<Node> n, const InfoHash& infohash, con
     msgpack::packer<msgpack::sbuffer> pk(&buffer);
     pk.pack_map(5+(network?1:0));
 
-    pk.pack(std::string("a")); pk.pack_map(3 +
-                               (query.where.getFilter() or not query.select.getSelection().empty() ? 1:0));
+    auto has_query = query.where.getFilter() or not query.select.getSelection().empty();
+    pk.pack(std::string("a")); pk.pack_map(3 + (has_query ? 1:0));
```
kaldoran: (has_query ? 1:0)  -> + has_query 
Since has_query is a boolean is will be convert into 1 or 0 (according to true / false)

include/opendht/node.h
sim590: - [x] reset auth_errors

src/network_engine.cpp
```diff
@@ -377,6 +377,9 @@ NetworkEngine::processMessage(const uint8_t *buf, size_t buflen, const SockAddr&
             break;
         }
         case MessageType::Reply:
+            if (msg.type == MessageType::AnnounceValue or msg.type == MessageType::GetValues)
```
aberaud: Listen ?

sim590: Indeed. Fixed.

Issue #135, opened 2016-09-22
OSX won't allow a larger datagram size than 9216 bytes


aberaud:
Ring account export wasn't working on OS X.
After investigation is seems related to:
https://github.com/BanTheRewind/Cinder-Asio/issues/9#issuecomment-67540675

aberaud:
Drop OS X support ?

sim590:
Since this issue is assigned to me, I pronounce the support for OSX now dropped.

aberaud:
The hidden message was: is your TCP patch ready soon ? :-P

PR #136, opened 2016-10-04
network: transmit large values in multiple packets

aberaud:
* add support to fragment values in smaller UDP packets
* NetworkEngine level changes
* backward compatible for small (<8KB) values


sim590:
This should address #135.

Issue #137, opened 2016-10-07
Dht: Detect connectivity changes


sim590:
When losing connection with the network and that there's no practical way getting notified by underlying network layers, we may ask the question: *How do you detect those changes so that the node can responsively recover?*

In particular, we can think of the event where one listens to some hash *h* and gets disconnected for a brief period long enough to flag a significative amount of nodes as expired. This list of nodes would particularly contain the closest nodes to the target hash *h*. Therefor, when getting connection back, we wouldn't allow the list of the closest to be sent messages from us. We would then be unsynced from any other nodes not suffering from such an eventuality.  For 10 minutes further *listen* requests would be sent to wrong nodes. The same is true for other operations.

We have analysed this concerning case and found a criteria for identifying this situation. The idea is to find the right amount of nodes flagged as expired indicating the anomaly. See Annex 1 for mathematical details.

[Annex 1](https://github.com/savoirfairelinux/opendht/files/516816/analysis.pdf)

sim590:
Here is a sample of node lists observed using the DHT nodes when suffering from the limitations so we can see the phenomenon.

[expired-nodes.txt](https://github.com/savoirfairelinux/opendht/files/528074/expired-nodes.txt)
[expired-nodes-after-reconnection.txt](https://github.com/savoirfairelinux/opendht/files/528075/expired-nodes-after-reconnection.txt)

PR #138, opened 2016-10-07
search: detect connectivity changes

sim590:
The number of bad nodes is limited in order to help determine presence of connectivity changes. This is intended to address #137.

According to the tables, 25 is a good average value for big networks. If the network is small, normal search expiration process will handle the situation.

I have made the choice of merging the events of an expired search and connectivity change event. I think this is clear that those two are strongly linked together if not actually the same case. @aberaud what do you think? Take a look at [this line](https://github.com/savoirfairelinux/opendht/pull/138/files#diff-96c057f80526cd0b3abd165dedb9f8cfR613)

### DIFF
--- a/include/opendht/dht.h
+++ b/include/opendht/dht.h
@@ -294,6 +294,16 @@ private:
        the target 8 turn out to be dead. */
     static constexpr unsigned SEARCH_NODES {14};
 
+    /* The number of bad nodes is limited in order to help determine
+     * presence of connectivity changes. See
+     * https://github.com/savoirfairelinux/opendht/issues/137 for details.
+     *
+     * According to the tables, 25 is a good average value for big networks. If
+     * the network is small, normal search expiration process will handle the
+     * situation.
+     * */
+    static constexpr unsigned SEARCH_MAX_BAD_NODES {25};
+
     /* Concurrent search nodes requested count */
     static constexpr unsigned MAX_REQUESTED_SEARCH_NODES {4};
 
--- a/src/dht.cpp
+++ b/src/dht.cpp
@@ -429,7 +429,7 @@ struct Dht::SearchNode {
      * Is this node expired or candidate
      */
     bool isBad() const {
-        return !node || node->isExpired() || candidate;
+        return not node or node->isExpired() or candidate;
     }
 };
 
@@ -553,7 +553,21 @@ struct Dht::Search {
     /**
      * @return The number of non-good search nodes.
      */
-    unsigned getNumberOfBadNodes() const;
+    unsigned getNumberOfBadNodes() const {
+        return std::count_if(nodes.begin(), nodes.end(), [](const SearchNode& sn) {
+            return sn.isBad();
+        });
+    }
+    unsigned getNumberOfConsecutiveBadNodes() const {
+        unsigned count = 0;
+        std::find_if(nodes.begin(), nodes.end(), [&count](const SearchNode& sn) {
+            if (not sn.isBad())
+                return true;
+            ++count;
+            return false;
+        });
+        return count;
+    }
 
     /**
      * Returns the time of the next "announce" event for this search,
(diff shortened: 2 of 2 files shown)


src/dht.cpp
```diff
@@ -1341,41 +1398,11 @@ Dht::searchStep(std::shared_ptr<Search> sr)
...
-                    if (not ait->permanent)
-                        ait = sr->announce.erase(ait);
-                    else
-                        ait++;
-                }
-                for (const auto& a : a_cbs)
-                    a(false, {});
-            }
-        }
+    if (sr->getNumberOfBadNodes() >= std::min(sr->nodes.size(), static_cast<size_t>(SEARCH_MAX_BAD_NODES))) {
```
aberaud: In the context of network with many natted nodes, purging bad nodes would cause a try-loop and the search would end up never be synced. Operations should keep going smoothly with as much as 7/8 unreachable nodes (BTW we should use that value to compute the number of nodes instead of 0.5 since it's the worst case scenario to support !)..
Search expiration should happen only when we believe that we are disconnected from the network, if there is no unexpired nodes left in the search (or if the first N nodes are expired since we try them in order), otherwise we should aim to remember bad nodes as much as possible to avoid trying them again and block the search.
The current method to check for expiration is probably well suited for our purpose, we only need to limit the maximum size of the search, and it will expire when every of its node is.

src/dht.cpp
```diff
@@ -575,7 +575,75 @@ struct Dht::Search {
...
+                if (++erased_count == count)
+                    break;
+            }
+        }
+        return erased_count > 0;
+    };
+
+    /**
+     * This method is called when we have discovered that the search may not be
+     * synced anymore. We have to
```
aberaud: "may not be synced anymore" -> "is expired"

include/opendht/dht.h
```diff
@@ -371,6 +381,7 @@ class Dht {
     Scheduler scheduler {};
     std::shared_ptr<Scheduler::Job> nextNodesConfirmation {};
     std::shared_ptr<Scheduler::Job> nextStorageMaintenance {};
+    std::shared_ptr<Scheduler::Job> nextConnectivityChange {};
```
aberaud: unused

src/dht.cpp
```diff
@@ -575,7 +575,75 @@ struct Dht::Search {
...
+     * NODE::NODE_EXPIRE_TIME minutes.
+     *
+     * @param now  The reference to now. Passing time_point::max() will remove
+     *             up to `count` expired nodes regardless of the time its been
+     *             expired.
+     * @param count  The number of expired nodes to remove.
+     *
+     * @return true if a node has been removed, else false.
+     */
+    bool removeExpiredNodes(time_point now, size_t count) {
```
aberaud: Honest question : why *remove* (not keep) up to count nodes ?

aberaud:
Agree search expiration to be considered as a connectivity loss.
Commented about a few things I found odd in the implementation.
Should we also call the dht conectvityChanged() as a result ?

src/dht.cpp
sim590: Reusing same code for [998](https://github.com/savoirfairelinux/opendht/pull/138/files#diff-96c057f80526cd0b3abd165dedb9f8cfR998) where we only want to remove 1 expired node.

aberaud: Yeah but why removing up to count nodes if a search expires ? For instance why not removing all expired nodes in that case ?

sim590: Indeed. My first impression was that using `SEARCH_MAX_BAD_NODES` would have done it, but now I see it doesn't.
- [ ] remove all bad nodes

src/dht.cpp
sim590: > BTW we should use that value to compute the number of nodes instead of 0.5 since it's the worst case scenario to support !

The number 25 is based on table 3 from the annex 1 of #137 (k=3/4). Going from 3/4 to 7/8, the number is at least 50. Which seems odd to me.

> Search expiration should happen only when we believe that we are disconnected from the network, if there is no unexpired nodes left in the search (or if the first N nodes are expired since we try them in order), otherwise we should aim to remember bad nodes as much as possible to avoid trying them again and block the search.

If I undestand correctly, you suggest to exclusively consider **expired** nodes in the node count, right? I also don't count consecutive expired nodes. May be that would be more appropriate.

- [ ] count consecutive **expired** nodes;

> The current method to check for expiration is probably well suited for our purpose, we only need to limit the maximum size of the search, and it will expire when every of its node is.

I'm not sure I get what you're saying. My understanding is that this patch should remove expired nodes from the affected `Search` before we reach the point where **all** nodes are expired. Hence the fact I replaced the expiration criteria to consider `min(nodes.size(), SEARCH_MAX_BAD_NODES)`. However, as you stated above, I should only consider expired nodes exclusively.

sim590:
I also note that we should do the following:
- [ ] call `connectivityChanged`;

src/dht.cpp
aberaud: > The number 25 is based on table 3 from the annex 1 of #137 (k=3/4). Going from 3/4 to 7/8, the number is at least 50. Which seems odd to me.

It makes sense, since if 7/8 nodes are expired, it's then normal to get see many consecutive unreachable nodes, so it takes more successive expired nodes to make sure we are actually disconnected and not just seeing random unreachable nodes.

> If I undestand correctly, you suggest to exclusively consider expired nodes in the node count, right? I also don't count consecutive expired nodes. May be that would be more appropriate.

Current implementation would consider a search with 8 good nodes, then many expired nodes as expired (but it's actually a fully working synced search).
We should count the number of consecutive expired nodes starting from the top of the search.

> I'm not sure I get what you're saying.

If I try to sum it up:
- we chose to consider search expiration as a connectivity loss
- When we detect network connectivity change, all searches should be affected by the expired nodes cleanup, (plus calling connectivityChanged). After this, the state of the DHT must be at best _CONNECTING_ to be consistent, and so that the connectivity change information bubbles up. 
- currently a search expires when all of its nodes are expired
- currently a search has no node count limit, (the good nodes could be anywhere in the node list), so the search expiration time is not bounded
- we can use our stat work results to put an upper limit on the search node count, and the current search expiration criteria can stay the same (it expires when all search nodes are expired). That would put an upper bound to the search expiration time.
- another possibility of implementation is to accept more nodes in the search, and count successive expired nodes, up to Max (our limit for search expiration). But then, what to do with the excessive nodes ? We won't be using them anyway, those nodes are far away, and just before expiring the search we started to send them requests (since they are next in the list after the Max expired nodes).

Yeah it sums it up, but I'm still not sure about how to handle some corner cases (listen recovery...).

sim590: > - we chose to consider search expiration as a connectivity loss

I might add that it's a connectivity loss with a subset of the network. It's possible we can still reach some nodes. For instance when losing connection only with outside our LAN.

> - another possibility of implementation is to not add a node count limit, and count successive expired nodes, up to Max (our limit for search expiration).

I have implemented this in 05e7ce6. Do you see any reason in favor of implementing the global search limit instead?

> -  But then, what to do with the excessive nodes ? We won't be using them anyway, those nodes are far away, and just before expiring the search we started to send them requests (since they are next in the list after the Max expired nodes).

I don't think it matters. Those nodes will expire and the same process will occur with nodes farther and farther from the search target subsequently used for the search, until all nodes in the cache are expired.

> Yeah it sums it up, but I'm still not sure about how to handle some corner cases (listen recovery...).

Won't the 'listen' operation recover when we are synced again? What do you have in mind?

Issue #81, opened 2016-06-27
Don't expose so many symbols


aviau:
It looks like OpenDHT exposes symbols that are not intended to be used by users.

See:

```
 dht::NetworkEngine::TransPrefix::GET_VALUES@Base 0.6.1
 dht::NetworkEngine::TransPrefix::ANNOUNCE_VALUES@Base 0.6.1
 dht::NetworkEngine::TransPrefix::PING@Base 0.6.1
 dht::NetworkEngine::TransPrefix::LISTEN@Base 0.6.1
 dht::NetworkEngine::TransPrefix::FIND_NODE@Base 0.6.1
 dht::NetworkEngine::bufferNodes(unsigned short, dht::InfoHash const&, std::vector<std::shared_ptr<dht::Node>, std::allocator<std::shared_ptr<dht::Node> > >&)@Base 0.6.1
 dht::NetworkEngine::bufferNodes(unsigned short, dht::InfoHash const&, signed char, std::vector<std::shared_ptr<dht::Node>, std::allocator<std::shared_ptr<dht::Node> > >&, std::vector<std::shared_ptr<dht::Node>, std::allocator<std::shared_ptr<dht::Node> > >&)@Base 0.6.1
```

From a quick discussion with another DD, it should be possible not to expose those symbols. This makes it easier to spot API breakage.

PR #140, opened 2016-10-14
hide symbols by default

aberaud:


include/opendht/def.h
```diff
@@ -0,0 +1,21 @@
+#pragma once
+
+#if defined _WIN32 || defined __CYGWIN__
+  #ifdef BUILDING_DLL
+    #ifdef __GNUC__
+      #define RING_PUBLIC __attribute__ ((dllexport))
```
sim590: Why define with the "RING" keyword? Why not use something that's always relevant to OpenDHT. I suggest changing the macro name to "DHT_PUBLIC" or "OPENDHT_PUBLIC".

aviau:
That would be great for #81

include/opendht/def.h
aberaud: done

PR #141, opened 2016-10-14
[dht] refresh operation for permanent put

sim590:
Since the introduction of the new "put" operation's behavior in 8f2cc93, the permanent put feature was broken every 2*Value::type.expiration minutes following a successful announce. This was due to the behavior where we didn't send a value if it was already announced to some nodes. However, since we reannounce with a time margin before a value expires, the reannounce operation would fail one time out of two.

### DIFF
--- a/include/opendht/network_engine.h
+++ b/include/opendht/network_engine.h
@@ -399,7 +399,7 @@ private:
     static const constexpr size_t NODE4_INFO_BUF_LEN {26};
     /* the length of a node info buffer in ipv6 format */
     static const constexpr size_t NODE6_INFO_BUF_LEN {38};
-    /* the maximum time we wait for an UDP reply before telling the link layer */
+    /* after a UDP reply, the period during which we tell the link layer about it */
     static constexpr std::chrono::seconds UDP_REPLY_TIME {15};
 
     /* Max. time to receive a full fragmented packet */
--- a/include/opendht/network_engine.h
+++ b/include/opendht/network_engine.h
@@ -49,6 +49,7 @@ public:
     // sent to another peer (http-like).
     static const constexpr uint16_t NON_AUTHORITATIVE_INFORMATION {203}; /* incomplete request packet. */
     static const constexpr uint16_t UNAUTHORIZED {401};                  /* wrong tokens. */
+    static const constexpr uint16_t NOT_FOUND {404};                     /* storage not found */
     // for internal use (custom).
     static const constexpr uint16_t INVALID_TID_SIZE {421};              /* id was truncated. */
     static const constexpr uint16_t UNKNOWN_TID {422};                   /* unknown tid */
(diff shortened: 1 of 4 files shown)


src/dht.cpp
```diff
@@ -1264,7 +1278,15 @@ void Dht::searchSendAnnounceValue(const std::shared_ptr<Search>& sr) {
                                                                     sn->token,
                                                                     onDone,
                                                                     onExpired);
-                                } else {
+                                } else if (hasValue and a.permanent) {
+                                    sn->acked[a.value->id] = network_engine.sendRefreshValue(sn->node,
```
sim590: - [ ] add log (warning) when sending 'refresh' request

src/dht.cpp
```diff
@@ -3292,6 +3316,28 @@ Dht::onAnnounce(std::shared_ptr<Node> node,
...
+
+    auto s = store.find(hash);
+    if (s != store.end()) {
+        DHT_LOG.DEBUG("[Storage %s] refreshed value (vid: %d).", hash.toString().c_str(), vid);
+        s->second.refresh(now, vid);
+    } else {
+        DHT_LOG.DEBUG("[node %s] got refresh value for unknown storage (id: %s).",
+                node->id.toString().c_str(),
+                hash.toString().c_str());
+        throw DhtProtocolException {DhtProtocolException::NOT_FOUND, DhtProtocolException::STORAGE_NOT_FOUND};
```
sim590: - [ ] Handle this error at reception.

src/network_engine.cpp
```diff
@@ -1071,6 +1089,59 @@ NetworkEngine::sendAnnounceValue(std::shared_ptr<Node> n, const InfoHash& infoha
...
+                const Blob& token,
+                RequestCb on_done,
+                RequestExpiredCb on_expired)
+{
+    auto tid = TransId {TransPrefix::REFRESH, getNewTid()};
+    msgpack::sbuffer buffer;
+    msgpack::packer<msgpack::sbuffer> pk(&buffer);
+    pk.pack_map(5+(network?1:0));
+
+    pk.pack(std::string("a")); pk.pack_map(3);
```
sim590: Token won't be packed with `pk.pack_map(3);`.
- [ ] properly pack token.

include/opendht/network_engine.h
```diff
@@ -375,7 +399,7 @@ class NetworkEngine final
     static const constexpr size_t NODE4_INFO_BUF_LEN {26};
     /* the length of a node info buffer in ipv6 format */
     static const constexpr size_t NODE6_INFO_BUF_LEN {38};
-    /* TODO */
+    /* the maximum time we wait for an UDP reply */
```
sim590: @aberaud: can you confirm on this commnt?

aberaud: from https://linux.die.net/man/2/sendto :

>  MSG_CONFIRM
>  -  Tell the link layer that forward progress happened: you got a successful reply from the other side. If the link layer doesn't get this it will regularly reprobe the neighbor (e.g., via a unicast ARP). Only valid on SOCK_DGRAM and SOCK_RAW sockets and currently only implemented for IPv4 and IPv6. See arp(7) for details. 

We set this flag if we got a message from the node in the last UDP_REPLY_TIME.

include/opendht/network_engine.h
```diff
@@ -399,7 +399,7 @@ class NetworkEngine final
     static const constexpr size_t NODE4_INFO_BUF_LEN {26};
     /* the length of a node info buffer in ipv6 format */
     static const constexpr size_t NODE6_INFO_BUF_LEN {38};
-    /* the maximum time we wait for an UDP reply */
+    /* the maximum time we wait for an UDP reply before telling the link layer */
```
aberaud: something like
/* after a UDP reply, the period during which we tell the link layer about it */

sim590:
I assume that it has been merged in c0612a1b.

PR #144, opened 2016-10-18
dht: uniformly sync SearchNodes reannouncing

sim590:
Historically, the next announce timing was based on the last acknowledgement received from a SearchNode. This behavior was invalid since the 'refresh' operation. In fact, for a given value that was "refreshed", its creation time would differ from the announcer's info. Therefor, all search node's reannouncing timings would differ, hence causing inconsistency on the network.


aberaud:
merged

PR #146, opened 2016-10-22
dht: make reannounce margin bigger, safer

sim590:
Reannouncing margin going from 5 seconds to ~~30~~10 seconds.

In the case of a "permanent put", refreshing values need to be done sooner than 5 seconds before the expiration time of the value. It is common for a "put" operation to take up to 5 seconds when a search is not synced. A search can easily become unsynced after 10 minutes.

### DIFF
--- a/include/opendht/dht.h
+++ b/include/opendht/dht.h
@@ -331,7 +331,7 @@ private:
     /* Timeout for listen */
     static constexpr std::chrono::seconds LISTEN_EXPIRE_TIME {30};
 
-    static constexpr std::chrono::seconds REANNOUNCE_MARGIN {5};
+    static constexpr std::chrono::seconds REANNOUNCE_MARGIN {10};
 
     static constexpr size_t TOKEN_SIZE {64};
 


sim590:
- Considering the scenario mentionned above and the normal observed behavior, the syncing process time should never take up to 30 seconds.

- The ideal would be to determine the maximum time syncing a search can take according that some nodes in the search are still synced. This value has to be lower than the time a search takes to sync from start. It's not rare to see a search take up to 5-8 seconds. Therefor, the time we're looking for should be lower than 5-8 seconds and the present value seems to meet the requirements. However, considering the following case where up to 8 nodes prior to a good node in the id-ordered list are expired, the syncing time may take up to 6 seconds (3 seconds per node, 4 nodes are queried concurently), then invalidating the present reannounce margin safety and next valid nodes in the id-ordered list could not respond successfuly to a "refresh" request, hence creating more traffic by forcing another "announce" request.

- As we have stated in #137, we can assume we can observe more than 20 consecutive unresponsive nodes in the id-oreded list before we consider the situation being a "disconnected" state. Therefor, 8 consecutive nodes in a same search not responding *may be a rare case*, but is still considered as possible.

Considering all of this, may be we could agree on 10 seconds margin instead of 30 or 5?

PR #149, opened 2016-10-26
network_engine: deserialize proper number of fields

aberaud:


src/network_engine.cpp
```diff
@@ -1329,14 +1329,15 @@ ParsedMessage::msgpack_unpack(msgpack::object msg)
         }
     } else if (auto raw_fields = findMapValue(req, "fields")) {
         if (auto rfields = findMapValue(*raw_fields, "f")) {
-            auto fields_ = rfields->as<std::set<Value::Field>>();
+            auto vfields = rfields->as<std::set<Value::Field>>();
             if (auto rvalues = findMapValue(*raw_fields, "v")) {
                 if (rvalues->type != msgpack::type::ARRAY)
                     throw msgpack::type_error();
-                for (size_t i = 0; i < rvalues->via.array.size; ++i) {
+                size_t val_num = rvalues->via.array.size / vfields.size();
```
sim590: Indeed. Fields are serialized as follows:
```
[f11,f12,...,f21,f22,...,fn1,fn2,...,fnm]
```
with `nm = rvalues->via.array.size`. But, the iterator `i` should be smaller than `n`, hence your change for
```
val_num = nm/m = n
```

aberaud:
merged as 6703e039b16d53360bf299f03104547538dc80b7

PR #152, opened 2016-11-05
Listen consistency

sim590:
Since the initial implementation of listen operation, multiple listen requests from a single node on a same hash was not possible without singular inconsistencies. The flaw was coming from the fact that nodes receiving listen request could not differentiate between listen refresh request and a new independent listen request.

This patch introduces the new `net::Socket` structure which in time will enable remote nodes to send update for a storage while differentiating between concurrent listen requests. A socket has a unique identifier which consists of a transaction id as defined in the OpenDHT protocol

### DIFF
--- a/include/opendht/network_engine.h
+++ b/include/opendht/network_engine.h
@@ -27,6 +27,7 @@
 #include "utils.h"
 #include "rng.h"
 #include "rate_limiter.h"
+#include "net.h"
 
 #include <vector>
 #include <string>
@@ -37,6 +38,9 @@
 #include <queue>
 
 namespace dht {
+namespace net {
+
+struct Request;
 
 #ifndef MSG_CONFIRM
 #define MSG_CONFIRM 0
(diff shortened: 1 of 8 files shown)


include/opendht/network_engine.h
```diff
@@ -293,67 +241,179 @@ class NetworkEngine final
...
+     * @param socket_id  The tid to use to write to the request socket.
+     * @param hash        The hash key of the value.
+     * @param want        Wether to send ipv4 and/or ipv6 nodes.
+     * @param ntoken      Listen security token.
+     * @param nodes       The ipv4 closest nodes.
+     * @param nodes6      The ipv6 closest nodes.
+     * @param values      The values to send.
      */
-    void tellListener(std::shared_ptr<Node> n, uint16_t rid, const InfoHash& hash, want_t want, const Blob& ntoken,
+    void tellListener(std::shared_ptr<Node> n, uint32_t socket_id, const InfoHash& hash, want_t want, const Blob& ntoken,
```
sim590: @aberaud. I was thinking about refactoring this part to a more generic code that would fit well with the new socket notion. I thought that, there could be an interface offered by the network engine equivalent to something like `NetworkEngine::write(Socket& socket)` but considering the fact that a unique function as such could not exist since it cannot know how to serialize the packets to write to the socket. I thought there could be some sort of model based on callbacks which would be defined at reception of the a request since such callbacks would be dependant to the request a node receives.

That being said, this patch as it is now is enough to be merged IMO. More enhancements could be added later.

src/network_engine.cpp
```diff
@@ -1260,6 +1312,9 @@ ParsedMessage::msgpack_unpack(msgpack::object msg)
         error_code = e->via.array.ptr[0].as<uint16_t>();
     }
 
+    if (auto t = findMapValue(req, "sid"))
```
sim590: Oh yes. I did leave it there by mistake.

src/dht.cpp
```diff
@@ -2201,16 +2207,20 @@ Dht::storageChanged(const InfoHash& id, Storage& st, ValueStorage& v)
...
+            auto f = l.second.query.where.getFilter();
+            if (f and not f(*v.data))
+                continue;
+            DHT_LOG.w(id, node_listeners.first->id, "[store %s] [node %s] sending update",
+                    id.toString().c_str(),
+                    node_listeners.first->toString().c_str());
+            std::vector<std::shared_ptr<Value>> vals {};
+            vals.push_back(v.data);
+            Blob ntoken = makeToken((const sockaddr*)&node_listeners.first->addr.first, false);
+            network_engine.tellListener(node_listeners.first, l.second.sid, id, 0, ntoken, {}, {},
```
sim590: Indeed!

sim590: done.

src/network_engine.cpp
```diff
@@ -435,19 +437,18 @@ NetworkEngine::process(std::unique_ptr<ParsedMessage>&& msg, const SockAddr& fro
     const auto& now = scheduler.time();
 
     if (msg->type == MessageType::Error or msg->type == MessageType::Reply) {
-        auto reqp = requests.find(msg->tid.getTid());
-        if (reqp == requests.end()) {
+        auto req_it = requests.find(msg->tid);
+        auto rsocket_it = opened_sockets.find(msg->tid);
```
sim590: done.

PR #153, opened 2016-11-07
opendht.pc.in: Link with gnutls and nettle

aviau:


aberaud:
Use pkg-config --static

PR #155, opened 2016-11-09
Multi-Arch support

aviau:
This patch was written by Gianfranco Costamagna as part of:
 - https://bugs.debian.org/843788


LocutusOfBorg:
GNUInstallDir is a cmake helper that sets the directory following the standard GNU conventions:
e.g. in this case sets LIBDIR, that in case of a Debian-based system defaults to the correct multiarch location.
If you want to know more about this helper, you can look at the documentation
https://cmake.org/cmake/help/v3.0/module/GNUInstallDirs.html

and the source code
https://sources.debian.net/src/cmake/3.6.2-2/Modules/GNUInstallDirs.cmake/

(look for debian_version)

thanks @aviau for the pull request!

PR #156, opened 2016-11-15
autoconf: robustly check for cython3

sim590:
On some platforms, cython3 binary is called `cython` and on others, it's called `cython3`. The usual way to check if a python module is by doing the following:
```sh
echo 'import module' | $PYTHON
# $? -eq 0: module is installed
# $? -neq 0: module isn't installed
```
where `$PYTHON` is the Python binary you wish to check. Cython is a module.

CQFD


sim590:
- [ ] fix inconsistency with previous code `$CYTHON` variable.

PR #158, opened 2016-11-22
storage: update listener after value refresh

sim590:
All values on OpenDHT live for a given period of time. If a value is put on a hash on which a node listens, the listener should be notified.

If one implements "permanent" put manually, listeners for that hash would be updated each 10 minutes, which is logical.

However, if one uses the "permanent put" feature implemented by OpenDHT, it won't get notified since a "refresh" doesn't trigger a `tellListener` call and I think it should.

For instance, this [patch](https://gerrit-ring.savoirfairelinux.com/#/c/5127/) for the Ring daemon to get a buddy notification would need this. Otherwise, all buddies would not appear online more than 10 minutes.

### DIFF
--- a/src/dht.cpp
+++ b/src/dht.cpp
@@ -189,13 +189,13 @@ struct Dht::Storage {
      * @param vid  The value id
      * @return true if a value storage was updated, false otherwise
      */
-    bool refresh(const time_point& now, const Value::Id& vid) {
+    ValueStorage* refresh(const time_point& now, const Value::Id& vid) {
         for (auto& vs : values)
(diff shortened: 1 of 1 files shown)


aberaud:
Refresh doesn't modify the value in any way, so I think it shouldn't inform listeners.

Listeners will always get the value initially and should drop duplicates anyway. OpenDHT users should never receive the same value multiple times in a listen callback.

The current proper way to fix the presence issue would be to perform a new get every 10 minutes. Otherwise we would need a robust OpenDHT API to know when a value is dropped from the network.

PR #159, opened 2016-11-22
Log functions are moved to the log.cpp

imadari:
Log functions are moved to the log.cpp in order to avoid multiple definitions of the logger functions.


sim590:
Yes. In fact, I was having trouble with that on [dpaste](https://github.com/sim590/dpaste/).

aberaud:
Thanks !
merged in 3724670fcce4c906ec437380d3708cc446180f51

PR #161, opened 2016-12-06
remove InfoHash::toCString()

yomgui1:
Implementation of InfoHash::toCString() is broken


sim590:
Just curious. Can you explain what's broken about it? Is the returned pointer dangling since it's a pointer to some instance of `std::string` which is deleted after the function returns?

PR #168, opened 2017-01-20
Announce op fixes

sim590:


src/dht.cpp
```diff
@@ -1591,20 +1591,20 @@ Dht::Search::getUpdateTime(time_point now) const
...
             continue;
-        if (not n.isAnnounced(id, now))
+        if (not n.isAnnounced(id))
             return false;
         if (++i == TARGET_NODES)
-            break;
+            return true;
     }
-    return i;
+    return i >= nodes.size() - getNumberOfBadNodes();
```
aberaud: isn't this already checked by the above loop ?

sim590: Good point. I'll remove this part.

src/dht.cpp
```diff
@@ -1591,20 +1591,20 @@ Dht::Search::getUpdateTime(time_point now) const
...
             continue;
-        if (not n.isAnnounced(id, now))
+        if (not n.isAnnounced(id))
             return false;
         if (++i == TARGET_NODES)
-            break;
+            return true;
     }
-    return i;
+    return false;
```
sim590: Dang. I did cover this with my last version of the condition. I got confused removing it. Let me do that appropriately.

PR #175, opened 2017-02-28
argon2: use as submodule

sim590:
I have made sure we use `ref.c` from argon2 as it is the case right now in master.


sim590:
**N.B**: this patch also bumps the source to the latest master.

sim590:
- [ ] add CMake and Autotools checks that argon2 repository is initialized.
- [ ] notice the procedure in README.md (`git submodule update --init`).

aberaud:
The goal of including argon2 is to make life easier for users while argon2 is progressively added to distributions. Since argon2 is now in Debian and soon Ubuntu, I added support to use the system library instead (used by default, with a fallback on the included version).
- [ ] this PR makes the build depend on internal file structure of argon2 so it should specify a commit (if argon2 file structure changes, OpenDHT build would break).
- [ ] make sure it's compatible as a fallback if the system library is not found (with include <argon2.h>). 
- [ ] can we init the submodule automatically only if pkgconfig doesn't find the system argon2 ? that would be the more convenient for users. The more we move on, more users will have an argon2 system library ; we shouldn't tell most users to init submodules if they don't have to.
To be honest IMHO the current way is simpler - just keeping a copy that we update - for the next few months, until we can simply drop the included version. Latest commits ensure people with a system version can use it.

sim590:
Hi @aberaud,

> this PR makes the build depend on internal file structure of argon2 so it should specify a commit (if argon2 file structure changes, OpenDHT build would break).

The submodule itself keeps track of the commit to be checked out. If you clone the OpenDHT repository and enter `git submodule update --init`, you'll end up with the same HEAD. When commited, a submodule is binded to a commit, so it won't move and break without us moving it.

> can we init the submodule automatically only if pkgconfig doesn't find the system argon2 ?

I guess we could do `git submodule init` automatically, but not `git submodule update --init` as that could be too intrusive in the user's workflow I imagine. Tell me if you think otherwise.

> To be honest IMHO the current way is simpler - just keeping a copy that we update - for the next few months, until we can simply drop the included version.

Doing so prevents us to know what commit was used and it's simply harder to maintain if we have to. Plus, my patch is ready for the most part, so... It's just better to merge it.

> Latest commits ensure people with a system version can use it.

That is not the issue. The issue is, as I said above, that the devs (us) cannot know what revision was used. It is troublesome when comes the time to choose another version because if, for instance, upstream says it's better. Note also that the submodule uses the latest master (as of when I uploaded the patch). To untar an archive into the repository is not guaranteeing the latest version for people to use it more than using a submodule. Actually, submodule is better here as I said.

I'll address the points listed according to my comments in the following hours/day.

sim590:
> I guess we could do git submodule init automatically, but not git submodule update --init as that could be too intrusive in the user's workflow I imagine. Tell me if you think otherwise.

Actually, it doesn't make sense what I said there. `update --init` has to be run ultimately if the submodule is needed.  Also, I guess it's fine if we automatically `update --init`.

sim590:
> can we init the submodule automatically only if pkgconfig doesn't find the system argon2 ?

@aberaud: What if someone uses `make dist` to generate a tar ball and that `git submodule update --init` is run from within the `configure` file? If it was to be automated, it would have to be written inside `autogen.sh`, then the submodule wil always be updated and init, but I don't mind this. I'm okay witth that. Are you?

sim590:
@aberaud: I have addressed your list of concerns. Tell me what you think.

sim590:
> The most important issue is that argon2 submodule headers are included (with cmake and autotools) and library built and linked (with autotools) even if the system argon2 is installed and supposedly used.

What do you mean by "included"? If you only mean that it should build with the appropriate headers specified by the platform, then yes it will as per this [line][cmake-line], this [line][autotools-line1] and this [line][autotools-line2].

[cmake-line]: https://github.com/savoirfairelinux/opendht/pull/175/files#diff-af3b638bc2a3e6c650974192a53c7291R125
[autotools-line1]: https://github.com/savoirfairelinux/opendht/pull/175/files#diff-af3b638bc2a3e6c650974192a53c7291R125
[autotools-line2]: https://github.com/savoirfairelinux/opendht/pull/175/files#diff-480477e89f9b6ddafb30c4383dcdd705R4

> Also make sure that we can properly fallback to the included argon2 if the system argon2 is not found. The only case where it should fail is with autotools, if --with-argon2 is passed and system argon is not found.

Then, you don't want to make the user have to pass the `--without-argon2` flag himself? I think your change in 358d8bd7b3001fc2c5c541324f3f1cfb4983c86e already introduced this behavior. I have tested on my machine and it works well on either build systems. In my change, I have used what you wrote when I reabsed my PR.

CMakeLists.txt
```diff
@@ -118,6 +122,7 @@ configure_file (
 
 include_directories (
     ./
+    argon2/include/
```
aberaud: Always includes local argon2 headers even system argon2 is used

CMakeLists.txt
```diff
@@ -34,6 +34,10 @@ if (NOT OPENDHT_ARGON2)
         message("Argon2 not found, using included version.")
         set(OPENDHT_ARGON2 ON)
     endif()
+else()
+    # make sure argon2 submodule is up to date and initialized
+    message("Initializing Argon2 submodule")
```
aberaud: Not run in case of fallback (line 34)

autogen.sh
```diff
@@ -1 +1,2 @@
+git submodule update --init
```
aberaud: run even if system argon2 is used

src/Makefile.am
```diff
@@ -51,3 +53,27 @@ nobase_include_HEADERS = \
 
 nobase_noinst_HEADERS = \
         ../include/opendht/request.h
+
+######################
+#  ARGON2 submodule  #
+######################
+
+noinst_LTLIBRARIES = libargon2.la
```
aberaud: - built even if system argon2 is used (check with WITH_INCLUDED_ARGON2 macro)
- dependency issue: bug can be reproduced easily without parallel build: `make clean` and `make` (_without_ `-j`):

> ../libtool: line 7485: cd: ../argon2/src/.libs: No such file or directory
libtool:   error: cannot determine absolute directory name of '../argon2/src/.libs'

tools/Makefile.am
```diff
@@ -4,10 +4,10 @@ noinst_HEADERS = tools_common.h
 AM_CPPFLAGS = -I../include
 
 dhtnode_SOURCES = dhtnode.cpp
-dhtnode_LDFLAGS = -lopendht -lreadline -L../src/.libs @GnuTLS_LIBS@
+dhtnode_LDFLAGS = -lopendht -lreadline -L@top_builddir@/src/.libs -L@top_builddir@/argon2/src/.libs @GnuTLS_LIBS@
```
aberaud: local argon2 path included even when system library is used. might use @argon2_LIBS@ ?

CMakeLists.txt
sim590: What's the purpose of this?

tools/Makefile.am
sim590: Yes. I forgot about the tools in my last patch. Thanks.

autogen.sh
sim590: > I suggest to run only of if argon2 directory doesn't exist.

There is a case where the *directory* doesn't exist **and** system argon2 submodule should be used. It seems like the second sentence doesn't answer the issue in the former. Anyway, checking if directory exists doesn't change anything since `git` will do that for us.

If Instead, we were to move the init call into the `configure.ac` spec, then the call would be part of the `configure` script. Therefore, the `git` call would occur even when run from a distributed tarball where the git repository wouldn't exist. As a quick example:
```sh
$ make dist
$ tar -xzf opendht-${version}.tar.gz
$ cd opendht-${version} && ./configure
```
The last line would give error. That's the reason why I decided to put the call into `autogen.sh` since it is a shell command that belongs into the repository scripts and not in the distributed scripts as I said earlier in other comments. This is a valid point since it is part of the features of autotools which we support. One could ask "what happens if the `autogen.sh` script is run outside of the repository" to which I would answer: "this is not a supported case since it's not handled by autotools and such case would only happen if the user would pull out the files from the repository which is not the way opendht is intended to be built."

CMakeLists.txt
sim590: Fixed.

tools/Makefile.am
sim590: FIxed.

src/Makefile.am
sim590: > - built even if system argon2 is used (check with WITH_INCLUDED_ARGON2 macro)

Still on it. I'll update when done.

> - dependency issue: bug can be reproduced easily without parallel build: make clean and make (without -j)
>> ../libtool: line 7485: cd: ../argon2/src/.libs: No such file or directory
>> libtool: error: cannot determine absolute directory name of '../argon2/src/.libs'

I don't experience this on 4646756. Logically, this should not happen because of the first point you mentionned. The path `../argon2/src/.libs` should always be found. Are you sure you're on the good revision?

sim590: I guess there may have a `src/libargon2.la` file lying in your directory. This is due to the switch between system and included argon2. To help preventing this, I'll add a `clean-local` rule.

sim590: > I don't experience this on 4646756.

I have just reproduced it. I'm on it.

CMakeLists.txt
aberaud: Indeed. Now that you mention it, I see it wasn't so clear.

autogen.sh
aberaud: what about putting the .gitmodule file in the dist archive ?

src/Makefile.am
sim590: Fixed.

sim590:
Now I have a final problem:
```
/bin/bash ../libtool  --tag=CXX   --mode=link g++  -O3 -pedantic-errors -DMSGPACK_DISABLE_LEGACY_NIL -DMSGPACK_DISABLE_LEGACY_CONVERT -lpthread -L/home/simon/prog/opendht/argon2/src/.libs -lpthread -o libopendht.la -rpath /usr/local/lib libopendht_la-dht.lo libopendht_la-node_cache.lo libopendht_la-callbacks.lo libopendht_la-routing_table.lo libopendht_la-network_engine.lo libopendht_la-utils.lo libopendht_la-infohash.lo libopendht_la-node.lo libopendht_la-value.lo libopendht_la-crypto.lo libopendht_la-securedht.lo libopendht_la-dhtrunner.lo libopendht_la-default_types.lo indexation/libopendht_la-pht.lo libopendht_la-log.lo  libargon2.la -lgnutls -lnettle
libtool:   error: cannot find the library 'libargon2.la' or unhandled argument 'libargon2.la'
```

So, from what I can understand, this is due to those three lines in `Makefile.am`:
```make
lib_LTLIBRARIES = libopendht.la
libopendht_la_LIBADD = @Argon2_LIBS@ @GnuTLS_LIBS@ @Nettle_LIBS@
noinst_LTLIBRARIES = libargon2.la
```
It seems like `*_LIBADD` doesn't specify dependencies or an order in which to build things. So, libtool tries to link with `libargon2.la` which has not been built yet. It thinks it can build it after building `libopendht.so`. In master, we go around the issue litterally by using subdirs makefiles. Since argon2 subdir is specified before src, the problem doesn't occur. However, there should be a way to specify the dependency. I'm still looking for it. The only thing that resolves this issue for now is using `lib_LTLIBRARIES = libargon2.la libopendht.la`, which is not suited since we don't want to install libargon2.

@aberaud: Do you have a suggestion?

CMakeLists.txt
```diff
@@ -116,12 +121,15 @@ configure_file (
...
 )
 
-include_directories (
+include_directories(
     ./
     include/
     include/opendht/
     ${CMAKE_CURRENT_BINARY_DIR}/include/
 )
+if(OPENDHT_ARGON2)
```
aberaud: group with other if(OPENDHT_ARGON2) directives here

autogen.sh
sim590: The `.git` directory wouldn't even be included in the dist archive since it is not known by autotools. Therefore, `git` commands wouldn't work.

sim590:
Thanks to @yomgui1 and a quite long search before, I have found:
```automake
noinst_LTLIBRARIES = libargon2.la
libopendht_la_DEPENDENCIES = libargon2.la
```
which solves the dependency issue. This is actually needed and should have been done from the begining (in master). However, as I said, we go around the problem by using the subdirs order in the main `Makefile.am` file.

CMakeLists.txt
sim590: Fixed.

Issue #177, opened 2017-03-24
[benchmark] enable loop interface tests


sim590:
The pyroute2 dependency makes tests unusable on platforms other than linux. Also, the project introduces [bugs](https://github.com/iovisor/bcc/issues/553) between versions.

Since OpenDHT permits loop interface usage, I suggest setting `pyroute2` as optional and permit loop interface tests mode in benchmark.

ghenry:
Still a requirement?  

Thanks.

PR #179, opened 2017-03-28
benchmark: no virtual namespaces by default

sim590:
Following the suggestion from #177, I have enabled the benchmark to work on interfaces other than virtual networks. To run tests on virtual namespaces, one should add the options `--virt-ns` and `--ifname some_name` since the default value for `--ifname` is now `lo`: the loop interface.

### DIFF
--- a/python/tools/benchmark.py
+++ b/python/tools/benchmark.py
@@ -40,8 +40,9 @@ class WorkBench():
     nodes and cluster to create, etc. This class is also used to initialise and
     finish the network.
     """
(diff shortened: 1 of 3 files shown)


sim590:
768bfa5 and fccd569 seem to give some conflict as we implemented the same thing :wink:

sim590:
Now, this PR cannot be merged yet. I have to take into account the change in `DhtRunner.run` signature. More precisely: use `SockAddr`.

sim590:

PR #182, opened 2017-03-29
fixes the contrib paths for the Visual Studio build

atraczyk:
- changes to the contribs to fixed commits
- supplies a batch file to fetch and patch the contrib
- makes better filters for the opendht project
- fixes the include path for argon2


aberaud:
still not building :-(

PR #183, opened 2017-04-03
pht: Fix wrong output from Prefix::toString

sim590:
`dht::indexation::blobToString` had been written from older code. Therefore, minor refactoring errors led to wrong output from the function. This PR fixes those errors and also limits the output of `Prefix::toString` to the bits of `Prefix::content_`. An additional method for bits of `Prefix::flags_` is now provided as `Prefix::flagsToString`.

### DIFF
--- a/src/indexation/pht.cpp
+++ b/src/indexation/pht.cpp
@@ -44,20 +44,7 @@ static std::string blobToString(const Blob &bl) {
 
     return ss.str();
 }
-
-std::string Prefix::toString() const {
-    std::stringstream ss;
-
-    ss << "Prefix : " << std::endl << "\tContent_ : \"";
-    ss << blobToString(content_);
-    ss << "\"" << std::endl;
-
-    ss << "\tFlags_   : \"";
-    ss << blobToString(flags_);
-    ss << "\"" << std::endl;
-
-    return ss.str();
-}
+std::string Prefix::toString() const { return blobToString(content_); }
 
 void Pht::Cache::insert(const Prefix& p) {
     size_t i = 0;
--- a/src/indexation/pht.cpp
+++ b/src/indexation/pht.cpp
@@ -48,6 +48,7 @@ std::string Prefix::firstBitsToString(const Blob &bl, size_t len) const {
 }
 
 std::string Prefix::toString() const { return firstBitsToString(content_, size_); }
+std::string Prefix::flagsToString() const { return firstBitsToString(flags_, size_); }
 
 void Pht::Cache::insert(const Prefix& p) {
     size_t i = 0;
--- a/src/indexation/pht.cpp
+++ b/src/indexation/pht.cpp
@@ -25,26 +25,29 @@ namespace dht {
 namespace indexation {
 
 /**
- * Output the blob into string and readable way
+ * Get bit string representation of a blob for a specified length.
  *
- * @param bl   : Blob to print
+ * @param bl   A Blob
+ * @param len  The number of bits to consider.
  *
- * @return string that represent the blob into a readable way
+ * @return string that represent the blob
  */
-static std::string blobToString(const Blob &bl) {
+std::string Prefix::firstBitsToString(const Blob &bl, size_t len) const {
+    if (len >= bl.size() * 8)
+        throw std::out_of_range("specified length larger than blob size");
     std::stringstream ss;
-    auto bn = bl.size() % 8;
-    auto n = bl.size() / 8;
+    auto bn = len % 8;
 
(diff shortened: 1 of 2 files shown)


src/indexation/pht.cpp
```diff
@@ -27,37 +27,27 @@ namespace indexation {
...
  *
- * @param bl   : Blob to print
+ * @param bl   A Blob
+ * @param nbr  The number of bits to consider.
  *
- * @return string that represent the blob into a readable way
+ * @return string that represent the blob
  */
-static std::string blobToString(const Blob &bl) {
+static std::string blobToString(const Blob &bl, size_t nbr = 0) {
```
sim590: > method name is not accurate. It doesn't convert a Blob to a string.

May be `firstBitsToString`?

> nbr should not be optional (we manipulate bits, not bytes), allowing to simplify line 37. One could argue nbr is not a good argument name (type size_t already tells it's a number).

I agree.

> optional, at your preference: might be private, non-static in Prefix, using size_ directly (use and design are specific to Prefix).

If you look at the historic of the files `pht.cpp` and `pht.h`, you'll see that it was first the case (see 2d8fe502f29aaddced8584283f2db370e9ee1a85). Not much is said as to why it was changed. I guess that we found that the function was not depending on the state of the instance of Prefix and could apply to other objects directly like `Blob`. However, I think that we can limit ourselves to the `Prefix` class for our needs as of now, like you suggest.

src/indexation/pht.cpp
```diff
@@ -25,39 +25,30 @@ namespace dht {
...
  *
- * @param bl   : Blob to print
+ * @param bl   A Blob
+ * @param len  The number of bits to consider.
  *
- * @return string that represent the blob into a readable way
+ * @return string that represent the blob
  */
-static std::string blobToString(const Blob &bl) {
+std::string Prefix::firstBitsToString(const Blob &bl, size_t len) const {
```
aberaud: If you choose to make it a non-static method, you should remove the second argument and checks, and use size_ directly (otherwise shall remain static).

aberaud:
Manipulating bits one-by-one is not common in IT, we usually manipulate bytes (groups of 8 bits). Prefix exists to do this specific job and bitsToString is the method to print a Prefix.
So bitsToString is not a generic method related to Blob but a specific method related to Prefix. For this reason I think it should be part of Prefix and not exposed elsewhere.

src/indexation/pht.cpp
sim590: @aberaud: In the case of `toString`, the ouput is the bit string up to `size_` bits since this is the most expected behavior. However, one could like to get the whole bit string of the prefix for debugging purposes. To be honest, this function is mostly being used in debugging purposes. That's why I left it this way.

sim590:
> Manipulating bits one-by-one is not common in IT, we usually manipulate bytes (groups of 8 bits). Prefix exists to represent a serie of bits and bitsToString is the method to print content or mask of a Prefix. I suggest "bitsToString" because it doesn't print the "first" bits, it prints all the bits in the prefix.

If you want to argue on this, I'll ask you then: "What is the data of which the `Prefix` class is the prefix of?" The answer is the `Key` under which is indexed an element. Then, in word theory, a word `w` of length `n` has `w_0` as prefix, but also `w_1`, `w_2`, up to `w_n = w` where `w_i` is the first letters of the word `w` up to the `i`th. Therefore, you understand that an instance of the `Prefix` class doesn't represent only one prefix, but all the prefixes of a certain key. Limiting `toString` method to the "current" prefix represented (when `size_` is considered at a given time) is fine as this is the most expected behavior. However, the private method which I name `firstBitsToString` could be useful to other extents than only yielding the bits up to `size_`. In other words, it can be extended to be used not only on the given state of the instance, but also on another state that it could extend to (by simply changing value of `size_`).

> Using a Blob that might have additional bits for data storage is an implementation detail.

I don't think that it's an implementation "detail". It's an important information. It contains the whole key for which we yield all prefixes.

> So bitsToString is not a generic method related to Blob but a specific method related to Prefix. For this reason I think it should be part of Prefix and not exposed elsewhere.

This is already the case in my last uploaded version.

Finally, keep in mind that another valuable information is part of `Prefix` for which `firstBitsToString` is required. I speak of `flags_` member. Therefore, here's another reason why `firstBitsToString` should keep its signature.

sim590:
@kaldoran, @aberaud, I haven't forgotten about this PR. In fact, I think about it everyday. I'll jump back on this as soon as I can.

sim590:

aberaud:
@sim590 Thanks I'll follow up on this PR

Issue #180, opened 2017-03-28
python: backward compatibility issue


sim590:
The whole benchmark is broken by 768bfa5d444bab566d7c0721200678d89bd226f1 since the `SockAddr` is not used by the benchmark and [no "legacy" signature (with address as string) is still available in the python API for bootstraping][legacy-bootstrap] :unamused:.

This is the case for benchmark, but also all python code which have already been written before 768bfa5d444bab566d7c0721200678d89bd226f1. For instance, the [pingpong][] script is broken. I think that some of ring's dht small swarm is using python bindings also. This is dangerous.

This kind of changes should be made more carefuly in the future since we never know who may use the python bindings also and we don't want to just drop support for API as fundamental as `DhtRunner.bootstrap`. I suggest, a legacy signature should be provided. If we want to deprecate it, then we could add a message on standard error specifying the matter, but anyhow I don't think it's good idea to remove this signature before a long time (couple of weeks or months).

[legacy-bootstrap]: https://github.com/savoirfairelinux/opendht/commit/768bfa5d444bab566d7c0721200678d89bd226f1#diff-46374efcb7905f7cd6ed62f3c1310dc6L287
[pingpong]: https://github.com/savoirfairelinux/opendht/blob/4540be7460d3760b32c2ffbe2b0826cd1da6314b/python/tools/pingpong.py#L30

aberaud:
it's a bug, this is exactly the kind of things that will be detected automatically with unit tests

sim590:
Oh. I just noticed that you in fact [kept the legacy signature available][legacy]. I didn't look close enough at first. So, yeah, like you say, this is a bug ;)

[legacy]: https://jenkins.ring.cx/job/daemon-ubuntu/384/consoleFull

aberaud:
Fixed by 99f8c2950b4c1d4af8e668e3d6ed317683924005

sim590:
@aberaud: But now, the bootstrap signature `bootstrap(string, string)` is not present anymore which was in part what I was complaining about. Don't you think that this canonical signature should always be available?

aberaud:
I think `bootstrap(string, string)` was always present (as you noticed above) and still is, but was not working because of a conflict with the now-removed signature.

sim590:
@aberaud. I really have trouble reading github's diffs.

PR #184, opened 2017-04-03
benchmark: fix pht insert test

sim590:
Output from `insertTest` in benchmark has been wrong since the begining due to minor errors. This PR fixes those errors and also **depends on #183** to build and run right. Keep that in mind before merging.

I have attached `before.txt` and `after.txt` files to compare outputs.

[before.txt](https://github.com/savoirfairelinux/opendht/files/889030/before.txt)
[after.txt](https://github.com/savoirfairelinux/opendht/files/889031/after.txt)

### DIFF
--- a/python/opendht_cpp.pxd
+++ b/python/opendht_cpp.pxd
@@ -187,7 +187,13 @@ cdef extern from "opendht/indexation/pht.h" namespace "dht::indexation":
     cdef cppclass Prefix:
         Prefix() except +
         Prefix(vector[uint8_t]) except +
+        Prefix(vector[uint8_t], size_t first) except +
+        Prefix(Prefix& p) except +
         string toString() const
+        string flagsToString() const
+        size_t size_
+        vector[uint8_t] flags_
+        vector[uint8_t] content_
     ctypedef pair[InfoHash, uint64_t] IndexValue "dht::indexation::Value"
     ctypedef map[string, vector[uint8_t]] IndexKey "dht::indexation::Pht::Key"
     ctypedef map[string, uint32_t] IndexKeySpec "dht::indexation::Pht::KeySpec"
--- a/python/tools/dht/tests.py
+++ b/python/tools/dht/tests.py
@@ -249,21 +249,22 @@ class PhtTest(FeatureTest):
     @staticmethod
     def lookupCb(vals, prefix):
         PhtTest.indexEntries = list(vals)
-        PhtTest.prefix = prefix.decode()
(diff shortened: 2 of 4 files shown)


sim590:
I also should mention that the test will fail to run properly due to the bug from #180. I have made sure the test works well on my end by reverting the SockAddr changes from the python layer.

python/tools/dht/tests.py
```diff
@@ -249,21 +249,22 @@ def _reset(self):
...
+        # DhtNetwork.Log.log('Index name: <todo>')
+        DhtNetwork.Log.log('Leaf prefix:', PhtTest.prefix)
         for v in vals:
-            DhtNetwork.log('[ENTRY]:', v)
+            DhtNetwork.Log.log('[ENTRY]:', v)
 
     @staticmethod
     def lookupDoneCb(ok):
-        DhtNetwork.log('[LOOKUP]:', PhtTest.key, "--", "success!" if ok else "Fail...")
+        DhtNetwork.Log.log('[LOOKUP]:', PhtTest.key, "--", "success!" if ok else "Fail...")
```
kaldoran: Log.log, looks like "a lot" of "Log"

python/opendht_cpp.pxd
```diff
@@ -187,7 +187,13 @@ cdef extern from "opendht/indexation/pht.h" namespace "dht::indexation":
     cdef cppclass Prefix:
         Prefix() except +
         Prefix(vector[uint8_t]) except +
+        Prefix(vector[uint8_t], size_t first) except +
+        Prefix(Prefix& p) except +
         string toString() const
+        string flagsToString() const
+        size_t size_
```
kaldoran: Totally agreed with that, since flags_ and content_ are internal variable, user shouldn't be able to edit them directly.
As @aberaud point out, there is a risk of inconsistence ( ou just mistake if you [try to] move around those bits )

python/tools/dht/tests.py
sim590: @kaldoran: What do you mean precisely? `Dhtnetwork.Log` is a class which has methods `log`, `warn` and `err`.

Do you suggest changing `Log.log` method to `Log.debug`?

kaldoran: Indeed, in order to avoid future mistake. [Log.debug or some other form]

sim590:

aberaud:
@sim590 Thanks I'll follow up on this PR

PR #185, opened 2017-04-03
pht: prevent unneeded use of shared_ptr

aberaud:
Prefer a const reference instead of shared_ptr if the shared_ptr
is not used and argument type is not changed in the method.

This allows to prevent useless shared_ptr allocations.

In getRealPrefix, use OpState to track state instead of using many
different shared_ptr.

### DIFF
--- a/include/opendht/indexation/pht.h
+++ b/include/opendht/indexation/pht.h
@@ -304,13 +304,13 @@ public:
     /* Specifications of the keys. It defines the number, the length and the
      * serialization order of fields. */
     using KeySpec = std::map<std::string, size_t>;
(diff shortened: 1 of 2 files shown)


sim590:
While I review this test, can you checkout [this branch](https://github.com/sim590/opendht/tree/pht-tests) and cherry-pick your commit to test that everything seems fine. You can try the `insert` test:
```sh
$ ./configure --enable-python && sudo make install
$ python3 benchmark.py --pht -t insert -o draw_trie -m 128 -n 128
```
You may experience put blocking as I think there is a bug in the underlying layer where some callbacks are not called after `put` operation. Nevertheless, try it and post the output of a successful test to see if output is consistent. The expected output should look like [this](https://github.com/savoirfairelinux/opendht/files/889031/after.txt).

PR #195, opened 2017-06-20
Avoid dereferencing an iterator to 'end()'

GrossoMoreira:
*h.end() only works on VS 2015 with
_SECURE_SCL=0
_ITERATOR_DEBUG_LEVEL=0

Also, if these flags are used when compiling opendht, it seems that any client program that wants to link against opendht also needs the same flags enabled, which is a terrible requirement to force on a client.

The proposed change makes the code a bit cleaner and doesn't require any preprocessor tricks.


aberaud:
thanks

Issue #218, opened 2017-11-08
Communicate with another node through a proxy server


AmarOk1412:
We need to be able to use the proxy server introduced in 4430b1e616490d84a2aa1ab682f54394a1942ce7 directly via the library.

The main goal is to avoid data consumption by stopping local synchronization when using the proxy.

Will be implemented in steps:

1. First, Implement a basic client which have the ability to get and put data.
2. Then, implement listen and connectivity changed behaviors.
3. Use it in the Ring project and some examples.

PR #219, opened 2017-11-09
proxy: implement basic DhtProxyClient with the ability to get and put.

AmarOk1412:
**WIP** cf 1. #218

+ Add an abstract class to describe the Dht
+ Now, we can modify the Dht linked to a SecureDht. So we can choose
between the classic Dht and the Dht using a proxy.
+ Implement a basic client which can Get and Put values on the Dht using
the proxy.

Note: Update dhtnode to have the ability to use the DhtProxyClient.
Todo: Implement Listen, ping, connectivityChanged for the client.

### DIFF
--- a/include/opendht/callbacks.h
+++ b/include/opendht/callbacks.h
@@ -56,6 +56,8 @@ struct OPENDHT_PUBLIC NodeStats {
      * Build a json object from a NodeStats
      */
     Json::Value toJson() const;
+    NodeStats() {};
+    NodeStats(const Json::Value& v);
 #endif //OPENDHT_PROXY_SERVER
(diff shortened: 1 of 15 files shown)


include/opendht/callbacks.h
```diff
@@ -56,6 +56,8 @@ struct OPENDHT_PUBLIC NodeStats {
      * Build a json object from a NodeStats
      */
     Json::Value toJson() const;
+    NodeStats() {};
```
aberaud: The bug comes from here. If you add a constructor, you must give default values (0) otherwise fields are not default-initialized

Issue #225, opened 2017-11-28
Supports Push notifications for mobile clients


AmarOk1412:
# Problem
On some platforms (like iOs), you can't really have a node running in background h24. We currently have the proxy, but we need to maintain the connection for `listening`operations.

# Solution
Have the possibility to supports push notifications on the proxy node.

# Warning
This option must be disabled by default. It **breaks** the distribution... because the proxy should have the certif for the app.

# How
Add `SUBSCRIBE`and `UNSUBSCRIBE` to `DhtProxyServer`. Subscribe will perform a Listen on the DHT and send values for the device (he must have a key) and return a token. Unsubscribe will do a cancelListen.

Because a client can not call `unsubscribe` we must have a timeout to remove old listen operations.

When the proxy server receives a value, it will send a push notification. To avoid tools like *Firebase*, we can host our own push notification server like https://github.com/appleboy/gorush. Sadly, it will use GCM or FCM for android and APN for iOs (if anyone know another viable solution)...

For Fdroid, this option must be disabled, and we will use listen like before (with or without DHT Proxy).

# Things to determine

1. Which alternative for push notification server? Currently gorush because it's under MIT License, looks like popular and active.
2. Subscribe and Unsubscribe should be done on the same node to works. So we have a problem if we use a load balancer. Maybe we will need a new message on the DHT to cancel 

![screenshot-2017-11-29 https mermaidjs github io 1](https://user-images.githubusercontent.com/1888628/33456024-0629fa8e-d5ec-11e7-9176-335fb66679a3.png)

AmarOk1412:
Ok.

What I developped is basically the same system described https://matrix.org/docs/spec/push_gateway/unstable.html. The only difference is we also pass encrypted values in the notification and they just use the notification as a wake-up notification. (We also use it has a wake up notification to enable the daemon and decrypt the data on the phone).

AmarOk1412:
![screenshot-2017-12-15 https mermaidjs github io](https://user-images.githubusercontent.com/1888628/34050442-86f8d244-e180-11e7-92ef-6a566f694ed6.png)

PR #228, opened 2017-12-12
Add support for push notifications

AmarOk1412:
cf #225

+ Add SUBSCRIBE and UNSUBSCRIBE for DhtProxyServer
+ Add the ability to stop Listen via a timeout (and send a push
Notification for the client)
+ Add the ability to use gorush to send push notifications
+ add the ability to inject push notifications in opendht to call
correct callbacks
+ if a timeout occurs and no cancelListen was called, relaunch the
listen operation
+ Pass devicekey to enable push notifications for DhtProxyClient
+ Update sockaddr to fix the build on all platforms

Note: Update dhtnode


src/dht_proxy_client.cpp
```diff
@@ -499,13 +533,33 @@ DhtProxyClient::cancelListen(const InfoHash&, size_t token)
...
                 return true;
+            } else {
+                // Just stop the request
+                if (listener.thread.joinable()) {
+                    // Close connection to stop listener?
+                    if (listener.req)
+                        restbed::Http::close(listener.req);
+                    listener.thread.join();
+                    listeners_.erase(it);
+                    lockListener_.unlock();
```
aberaud: not necessary: mutex is unlocked on destruction ok lock_guard

tools/dhtnode.cpp
```diff
@@ -101,13 +113,17 @@ void cmd_loop(std::shared_ptr<DhtRunner>& dht, dht_params& params)
 #if OPENDHT_PROXY_SERVER
```
aberaud: This up to line 128 should not be part of the cmd_loop, should go to main, to be respected even in non-interactive modes. Maybe make proxies a global, of pass reference to cmd_loop ?

tools/dhtnode.cpp
```diff
@@ -472,6 +510,15 @@ main(int argc, char **argv)
         }
 
         if (params.daemonize or params.service) {
+#if OPENDHT_PROXY_SERVER
+            if (params.proxyserver != 0) {
+                new DhtProxyServer(dht, params.proxyserver
```
aberaud: See comment above (also true for proxyclient). Here, the object lifetime is not correctly handled, so the server is not shutdown properly on exit.

tools/dhtnode.cpp
AmarOk1412: I prefer the reference.

PR #231, opened 2017-12-14
dht_interface: fix logger issues

### LINKED ISSUE #230: Cannot use -l or --logfile option when --daemonize or -d is used
Hello,

There doesn't appear to be a way to generate log output (either to file, syslog, or stdout) when using the --daemonize or -d option of the dhtnode program.

This works: ``dhtnode --syslog -b bootstrap.ring.cx`` or ``dhtnode -v -b bootstrap.ring.cx`` or ``dhtnode  --logfile=dhtnode.log -b bootstrap.ring.cx``

These args are accepted but appear to not have an effect (no logs in syslog and the logfile is not created) ``dhtnode --syslog --daemonize -b bootstrap.ring.cx`` or ``dhtnode --daemonize --logfile=dhtnode.log -b bootstrap.ring.cx``

AmarOk1412:
Make DHT_LOG static to use it in the whole library.

Fix #230

### DIFF
--- a/include/opendht/dht_interface.h
+++ b/include/opendht/dht_interface.h
@@ -214,7 +214,6 @@ public:
 
     virtual std::vector<SockAddr> getPublicAddress(sa_family_t family = 0) = 0;
 
-    Logger DHT_LOG;
     /**
      * Enable or disable logging of DHT internal messages
      */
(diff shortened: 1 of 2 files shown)


include/opendht/dht_interface.h
```diff
@@ -214,11 +214,11 @@ class OPENDHT_PUBLIC DhtInterface {
 
     virtual std::vector<SockAddr> getPublicAddress(sa_family_t family = 0) = 0;
 
-    Logger DHT_LOG;
+    static Logger DHT_LOG;
```
aberaud: the logger should be independent per node. For instance each node might log to a different file.

AmarOk1412: Oh. Indeed...
I think you already told me that. I'll do it asap. Just need to provide docs first. If I will not do it today, I'll do it next week should be easy to fix :)

Issue #238, opened 2018-02-08
dhtnode cannot encrypt data for the destination


kiddouk:
hey guys,

I tried to encrypt data between 2 dhtnodes and found out that this is not working at the moment. Here are the logs.

``` 
>> [080199.334204] Registering certificate for f026ad0393211631357a3f7c5897cda8f33eb328
[080199.334223] Found certificate for f026ad0393211631357a3f7c5897cda8f33eb328
[080199.334318] Encrypting data for PK: f026ad0393211631357a3f7c5897cda8f33eb328
[080199.335767] Error putting encrypted data: Must be an RSA key
```

After investigation, it looks like the certificate recorded in the DHT is an `GNUTLS_PK_EC` while the `encrypt` method expect to encrypt the data in RSA.

I will submit a PR for that.

PR #239, opened 2018-02-08
Generates RSA pairs for dhtnode instead of EC cert

kiddouk:
* dhtnode was generating twice an EC cert which rendered encryption
impossible since opendht lib is expecting an RSA public key to encrypt
the data for the destination.

* fix savoirfairelinux/opendht#238


aberaud:
Thanks

PR #248, opened 2018-02-27
build: make indexation optional

aberaud:


tools/dhtnode.cpp
```diff
@@ -253,6 +257,7 @@ void cmd_loop(std::shared_ptr<DhtRunner>& dht, dht_params& params
         dht::InfoHash id;
 
         if (op == "il" or op == "ii") {
+#ifdef OPENDHT_INDEXATION
```
AmarOk1412: Not mendatory, but I will prefer the if condition (op == "ii" and il) in the #ifdef to avoid any build warnings (useless if).

src/Makefile.am
```diff
@@ -64,6 +62,11 @@ libopendht_la_SOURCES += dht_proxy_client.cpp
 nobase_include_HEADERS += ../include/opendht/dht_proxy_client.h ../include/opendht/dht_interface.h
 endif
 
+if ENABLE_INDEXATION
```
AmarOk1412: OPENDHT_INDEXATION or change the previous add_definitions.

configure.ac
```diff
@@ -32,6 +32,13 @@ AS_IF([test "x$enable_logs" != "xno"], [
 	AC_DEFINE([OPENDHT_LOG], [false], [Define if DHT logs are enabled])
 ])
 
+dnl Check for indexation
+AC_ARG_ENABLE([logs], [AS_HELP_STRING([--enable-indexation], [Disble DHT logs])])
```
AmarOk1412: This line is incorrect.
By the way, there is some `Disble`

CMakeLists.txt
```diff
@@ -51,7 +52,7 @@ if (OPENDHT_PROXY_SERVER OR OPENDHT_PROXY_CLIENT)
 endif()
 
 # Build flags
-set (CMAKE_CXX_STANDARD 11)
+set (CMAKE_CXX_STANDARD 14)
```
AmarOk1412: We are not switching to c++14 yet.

AmarOk1412:
*By the way, this option should be added in the wiki*

tools/dhtnode.cpp
aberaud: done

src/Makefile.am
aberaud: this is an AM_CONDITIONAL not a define

configure.ac
aberaud: done

CMakeLists.txt
aberaud: done

src/Makefile.am
AmarOk1412: oh my bad... was in the CMakeLists...

PR #263, opened 2018-03-26
Cache: improve value expiration handling

aberaud:


src/dht.cpp
```diff
@@ -538,59 +604,7 @@ Dht::searchStep(Sp<Search> sr)
...
-                            if (auto sr = ws.lock()) {
-                                scheduler.edit(sr->nextSearchStep, scheduler.time());
-                                if (auto sn = sr->getNode(node))
-                                    sn->onValues(query, std::move(answer), types, scheduler.time());
-                                    //onGetValuesDone(node, answer, sr, query);
-                            }
-                        }
-                    );
-                }
+                searchSychedNodeListen(sr, n);
```
AmarOk1412: Synched?

PR #273, opened 2018-04-09
proxy server: keep only one push listener per token/key

aberaud:


src/dht_proxy_server.cpp
```diff
@@ -352,24 +352,26 @@ DhtProxyServer::subscribe(const std::shared_ptr<restbed::Session>& session)
...
+                    // New listener
                     pushListener->second.isAndroid = isAndroid;
 
                     // The listener is not found, so add it.
                     ++tokenPushNotif_;
                     token = tokenPushNotif_;
-                    Listener listener;
                     listener.token = token;
                     listener.internalToken = dht_->listen(infoHash,
                         [this, pushToken, token, isAndroid, clientId](std::vector<std::shared_ptr<Value>> /*value*/) {
```
AmarOk1412: We pass the token here by copy. but it's changed before!!!

src/dht_proxy_server.cpp
```diff
@@ -352,24 +352,26 @@ DhtProxyServer::subscribe(const std::shared_ptr<restbed::Session>& session)
...
-                                scheduler_.edit(listener.expireJob, scheduler_.time() + OP_TIMEOUT);
-                            }
-                            s->close(restbed::OK, "{\"token\": " + std::to_string(listener.token) + "}\n");
-                            schedulerCv_.notify_one();
-                            return;
+                    auto list = pushListener->second.listeners.emplace(infoHash, Listener{});
+                    auto& listener = list.first->second;
+                    if (not list.second) {
+                        // Refresh existing listener
+                        listener.token = tokenFromReq;
```
AmarOk1412: Warning, when we refresh this token, it replaces the previous one. But the previous one is still used in the push notifications because of the comment below.

PR #278, opened 2018-04-18
proxy client: close old socket when restarting listeners

aberaud:


src/dht_proxy_client.cpp
```diff
@@ -864,67 +864,70 @@ DhtProxyClient::restartListeners()
...
-                                            callbacks_.emplace_back([cb, value, okCb](){
-                                                okCb->set_value(cb({value}, false));
-                                            });
-                                            loopSignal_();
-                                        }
-                                        futureCb.wait();
-                                        if (!futureCb.get()) {
-                                            return;
-                                        }
+                                    if (not filter or filter(*value))  {
```
AmarOk1412: Just need to try that cb is not nullptr.

AmarOk1412:
Tests Ok

src/dht_proxy_client.cpp
aberaud: Done

PR #285, opened 2018-04-23
proxy: remove useless token for push listeners

AmarOk1412:
Since we authentificate listeners with the clientId and the hash and
we use only one listener by hash/clientId, we doesn't need to store
a token always equals to 0.

Note: also avoid to get multiple hashes for one push notification.


AmarOk1412:
Also fix the multiple get for each push :)

PR #289, opened 2018-05-06
proxy client: refactor

aberaud:
Factor various ways of sending listen requests.


AmarOk1412:
This PR don't build if only the proxy is enabled, I'll check after :)

(`'fillBodyToGetToken' was not declared in this scope`)

PR #290, opened 2018-05-07
search: announce immediately if synced and not pending/completed

aberaud:


src/request.h
```diff
@@ -57,6 +57,15 @@ struct Request {
     bool pending() const { return state_ == State::PENDING; }
     bool over() const { return not pending(); }
     State getState() const { return state_; }
+    char getStateChar() const {
+        switch (state_) {
+            case State::PENDING:   return 'f';
```
AmarOk1412: why f not p?

But this PR is ok for me

PR #292, opened 2018-05-11
proxy client: allow concurrent proxy info requests

aberaud:


src/dht_proxy_client.cpp
```diff
@@ -490,11 +497,11 @@ DhtProxyClient::getProxyInfos()
 
     // Try to contact the proxy and set the status to connected when done.
     // will change the connectivity status
-    if (statusThread_.joinable())
-        statusThread_.join();
-    statusThread_ = std::thread([this, serverHost]{
-        struct SuccessCount {std::atomic_uint ipv4 {0}, ipv6 {0};};
-        auto successCount = std::make_shared<SuccessCount>();
+    if (statusThread_.joinable()) {
+        statusThread_.detach();
```
AmarOk1412: Can be dangerous

PR #303, opened 2018-05-29
dht_proxy_server: avoid mutex when getting /

AmarOk1412:
dht_proxy_server: avoid mutex when getting /

The dht_proxy_server should be always up, so getNodesStats will
be pretty always up-to-date. We can add it into a job and avoid
to wait for this method to finish (and this method use an
internal lock) to have a quick answer about the status of the proxy.

Note: the update is forced during the initialization of the proxy

### DIFF
--- a/include/opendht/dht_proxy_server.h
+++ b/include/opendht/dht_proxy_server.h
@@ -21,12 +21,13 @@
 
 #pragma once
 
+#include "callbacks.h"
 #include "def.h"
-#include "sockaddr.h"
 #include "infohash.h"
-#include "scheduler.h"
-#include "value.h"
 #include "proxy.h"
+#include "scheduler.h"
+#include "sockaddr.h"
+#include "value.h"
 
 #include <thread>
 #include <memory>
@@ -237,6 +238,9 @@ private:
     std::thread schedulerThread_;
 
     Sp<Scheduler::Job> printStatsJob_;
+    mutable std::mutex statsMutex_;
+    mutable NodeStats ipv4Stats_ {};
+    mutable NodeStats ipv6Stats_ {};
 
     // Handle client quit for listen.
     // NOTE: can be simplified when we will supports restbed 5.0
--- a/src/dht_proxy_server.cpp
+++ b/src/dht_proxy_server.cpp
@@ -30,6 +30,8 @@
 #include <functional>
 #include <limits>
 
+#include <iostream>
+
 using namespace std::placeholders;
 
 namespace dht {
@@ -128,6 +130,16 @@ DhtProxyServer::DhtProxyServer(std::shared_ptr<DhtRunner> dht, in_port_t port , 
                 std::cout << getStats().toString() << std::endl;
                 scheduler_.edit(printStatsJob_, scheduler_.time() + PRINT_STATS_PERIOD);
             }
+            if (not stopListeners) {
+                // Refresh stats cache
+                auto newIpv4Stats = dht_->getNodesStats(AF_INET);
+                auto newIpv6Stats = dht_->getNodesStats(AF_INET6);
+                {
+                    std::lock_guard<std::mutex> lck(statsMutex_);
+                    ipv4Stats_ = std::move(newIpv4Stats);
+                    ipv6Stats_ = std::move(newIpv6Stats);
+                }
+            }
         });
         while (service_->is_up()  and not stopListeners) {
             std::unique_lock<std::mutex> lock(schedulerLock_);
@@ -205,8 +217,19 @@ DhtProxyServer::getNodeInfo(const std::shared_ptr<restbed::Session>& session) co
                     if (id)
                         result["id"] = id.toString();
                     result["node_id"] = dht_->getNodeId().toString();
-                    result["ipv4"] = dht_->getNodesStats(AF_INET).toJson();
-                    result["ipv6"] = dht_->getNodesStats(AF_INET6).toJson();
+                    {
+                        std::lock_guard<std::mutex> lck(statsMutex_);
+                        if (ipv4Stats_.good_nodes == 0 && ipv6Stats_.good_nodes == 0) {
+                            // NOTE: we want to avoid the disconnected state as much as possible
+                            // So, if the node is disconnected, we should force the update of the cache
+                            // and reconnect as soon as possible
+                            // This should not happen much
+                            ipv4Stats_ = dht_->getNodesStats(AF_INET);
+                            ipv6Stats_ = dht_->getNodesStats(AF_INET);
+                        }
+                        result["ipv4"] = ipv4Stats_.toJson();
+                        result["ipv6"] = ipv6Stats_.toJson();
+                    }
                     result["public_ip"] = s->get_origin(); // [ipv6:ipv4]:port or ipv4:port
                     Json::StreamWriterBuilder wbuilder;
                     wbuilder["commentStyle"] = "None";


include/opendht/dht_proxy_server.h
```diff
@@ -237,6 +238,9 @@ class OPENDHT_PUBLIC DhtProxyServer
     std::thread schedulerThread_;
 
     Sp<Scheduler::Job> printStatsJob_;
+    Sp<Scheduler::Job> getNodeStatsJob_;
+    mutable NodeStats ipv4Stats_ {};
```
aberaud: shouldn't be mutable

include/opendht/dht_proxy_server.h
```diff
@@ -237,6 +238,9 @@ class OPENDHT_PUBLIC DhtProxyServer
     std::thread schedulerThread_;
 
     Sp<Scheduler::Job> printStatsJob_;
+    Sp<Scheduler::Job> getNodeStatsJob_;
```
aberaud: merge with printStatsJob to ensure stats are printed after the update (and not maybe just before)

src/dht_proxy_server.cpp
```diff
@@ -205,8 +215,14 @@ DhtProxyServer::getNodeInfo(const std::shared_ptr<restbed::Session>& session) co
                     if (id)
                         result["id"] = id.toString();
                     result["node_id"] = dht_->getNodeId().toString();
-                    result["ipv4"] = dht_->getNodesStats(AF_INET).toJson();
-                    result["ipv6"] = dht_->getNodesStats(AF_INET6).toJson();
+                    if (ipv4Stats_.good_nodes == 0 && ipv6Stats_.good_nodes == 0) {
```
aberaud: * Shouldn't be needed (run the update on startup if required)
* Doesn't run on startup but every time there is no good known node which is not the intent

src/dht_proxy_server.cpp
```diff
@@ -129,6 +132,13 @@ DhtProxyServer::DhtProxyServer(std::shared_ptr<DhtRunner> dht, in_port_t port ,
                 scheduler_.edit(printStatsJob_, scheduler_.time() + PRINT_STATS_PERIOD);
             }
         });
+        getNodeStatsJob_ = scheduler_.add(scheduler_.time(), [this]{
+            if (service_->is_up() and not stopListeners) {
+                ipv4Stats_ = dht_->getNodesStats(AF_INET);
```
aberaud: not thread-safe but I guess that's ok since it's a POD and consistency is not a requirement for this data.

include/opendht/dht_proxy_server.h
AmarOk1412: I do this because I use this in getNodeInfo() which is const.

include/opendht/dht_proxy_server.h
AmarOk1412: printStatsJob_ print only subscribers and listeners, I don't think it's necessary? But yeah we can only use one job

src/dht_proxy_server.cpp
AmarOk1412: Not sure what you mean by "Doesn't run on startup but every time there is no good known node which is not the intent". Here I update the stats only if there is no good known node.

src/dht_proxy_server.cpp
AmarOk1412: dht_ should exists till the proxy exists. And the scheduler can only launch this job once at a time?

src/dht_proxy_server.cpp
aberaud: According to the comment the intent was to update on startup but that's not what the code does and there seems to be no reason to update only if there is no good node.

include/opendht/dht_proxy_server.h
aberaud: Indeed it's different stats. Might still optionally be merged in a single job.

src/dht_proxy_server.cpp
aberaud: dht_ is not the issue, consistency of ipv4Stats_ is but it doean't really matter

src/dht_proxy_server.cpp
AmarOk1412: In fact. If I update on startup, the dhtrunner can haven't discovered any node for now. Here I just want to be connected to the proxy as soon as possible (after dht.cpp discover some nodes in the network). I don't know if I'm clear enough?

src/dht_proxy_server.cpp
```diff
@@ -129,6 +132,13 @@ DhtProxyServer::DhtProxyServer(std::shared_ptr<DhtRunner> dht, in_port_t port ,
                 scheduler_.edit(printStatsJob_, scheduler_.time() + PRINT_STATS_PERIOD);
             }
         });
+        getNodeStatsJob_ = scheduler_.add(scheduler_.time(), [this]{
+            if (service_->is_up() and not stopListeners) {
```
aberaud: service doesn't need to be up, might this allow to update the stats during startup with this job ?

include/opendht/dht_proxy_server.h
aberaud: IMO there really is no reason to update from getNodeInfo() (the goal of this patch is to put the load on a separate job).

src/dht_proxy_server.cpp
AmarOk1412: Yeah the only issue I see is that we make a GET / at the same time the scheduler is called. Here we will have 2 write operations for the same value. In the other case the scheduler can't start two `getNodeStatsJob_` at the same time?

Anyway if we decide to keep the write operations in `getNodeInfo` I can add a lock_guard to be sure.

include/opendht/dht_proxy_server.h
AmarOk1412: the only case I see is that we set up a dhtproxyserver and we directly try to do a listen on it. (this is the case in `dhtproxytester.cpp`)

I can just disable unit tests on the proxy till I found a better solution to do that.

src/dht_proxy_server.cpp
AmarOk1412: Bad copy paste from the previous job. But indeed no need to check this here

src/dht_proxy_server.cpp
aberaud: Yes with your explanation actually the code makes sense, and that's a good reason to update if there is no good known node: the node might go through a disconnected state, on startup or later, and in that case we want to update the state ASAP. Update the comments to make it clear.

For a long-term solution we might have multiple dht state listeners and update the stats in case of connectivity change.

src/dht_proxy_server.cpp
aberaud: There also might be one or more read operations while data is written so the data might be inconsistent.

AmarOk1412: arf yeah I forgot that the write operation is not atomic :/

I'll add a lock to be sure

src/dht_proxy_server.cpp
AmarOk1412: Yeah the comment was unclear. I'll fix this tomorrow morning.

include/opendht/dht_proxy_server.h
AmarOk1412: Done

src/dht_proxy_server.cpp
AmarOk1412: Done

src/dht_proxy_server.cpp
AmarOk1412: Done

src/dht_proxy_server.cpp
AmarOk1412: Done

PR #311, opened 2018-06-18
dhtrunner: drop packet if processing took too long

aberaud:


AmarOk1412:
Looks good to me just to wait for the result on the big proxy..

PR #327, opened 2018-08-12
value_cache: expire value if expiration is now

aberaud:


AmarOk1412:
I think the best way to test this patch is to  bump the dht proxy :)

PR #346, opened 2018-11-10
dht_proxy_client: fix finished behavior

AmarOk1412:
callback_.emplace_back() returns immediately, so, in some case,
finished can be set to true even before we execute one callback.
Finished should be set to true only when donecb is executed or
if a callback returns false.

This improve the reactivity of the proxy client.

### DIFF
--- a/src/dht_proxy_client.cpp
+++ b/src/dht_proxy_client.cpp
@@ -254,7 +254,8 @@ DhtProxyClient::get(const InfoHash& key, GetCallback cb, DoneCallback donecb, Va
     o.thread = std::thread([=](){
         // Try to contact the proxy and set the status to connected when done.
         // will change the connectivity status
-        auto ok = std::make_shared<std::atomic_bool>(true);
+        struct GetState{ std::atomic_bool ok {true}; std::atomic_bool stop {false}; };
+        auto state = std::make_shared<GetState>();
         restbed::Http::async(req,
             [=](const std::shared_ptr<restbed::Request>& req,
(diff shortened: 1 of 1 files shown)


AmarOk1412:
@aberaud  A good way to reproduce the bug:

1. Use a dhtproxy with any Ring client
2. Spam someone with messages
3. Without this patch, some messages will fail even with you contact connected. With this patch, all messages will arrive.

src/dht_proxy_client.cpp
```diff
@@ -252,6 +252,7 @@ DhtProxyClient::get(const InfoHash& key, GetCallback cb, DoneCallback donecb, Va
     o.req = req;
     o.finished = finished;
     o.thread = std::thread([=](){
+        auto stop_cb = std::make_shared<std::atomic_bool>(false);
```
aberaud:
```suggestion
        struct GetState{ std::atomic_bool ok {true}; std::atomic_bool stop {false}; };
        auto state = std::make_shared<GetState>();
```

aberaud: line 266: `while (restbed::Http::is_open(req) and not *finished and not state->stop) {`
line 268: `if (*finished or state->stop)`

AmarOk1412: Done :) Indeed we can group these two options.

Note: Our travis build is broken because of these commit: https://github.com/savoirfairelinux/opendht/commit/b7b5152348091a44eea16da61be9fbfb09203992
Too old on the travis machine.

PR #348, opened 2018-11-22
dht: don't send probe query for non permanent puts

aberaud:


src/dht.cpp
```diff
@@ -471,25 +470,49 @@ void Dht::searchSendAnnounceValue(const Sp<Search>& sr) {
...
+            if (n.getAnnounceTime(a.value->id) <= now) {
+                if (a.permanent) {
+                    sendQuery = true;
+                } else {
+                    DHT_LOG.w(sr->id, n.node->id, "[search %s] [node %s] sending 'put' (vid: %d)",
+                            sr->id.toString().c_str(), n.node->toString().c_str(), a.value->id);
+                    n.acked[a.value->id] = std::make_pair(network_engine.sendAnnounceValue(n.node,
+                                                    sr->id,
+                                                    a.value,
+                                                    a.permanent ? time_point::max() : a.created,
```
AmarOk1412: Never true cause a.permanent is checked before.

PR #352, opened 2018-11-30
fixed a linking error in openwrt-18.06: undefined reference to 'tgetnum'

zhengfish:
Signed-off-by: zhengfish <zhengfish@gmail.com>


zhengfish:
Got it. Thanks.

PR #355, opened 2018-12-07
Optimizer UDP sending logic

hanyf:
It seems running good.

### DIFF
--- a/src/network_engine.cpp
+++ b/src/network_engine.cpp
@@ -290,9 +290,9 @@ NetworkEngine::requestStep(Sp<Request> sreq)
         node.setExpired();
         if (not node.id)
             requests.erase(req.tid);
+        if (req.on_expired)
+            req.on_expired(req, false);
         return;
-    } else if (req.attempt_count == 1) {
-        req.on_expired(req, false);
     }
 
     auto err = send((char*)req.msg.data(), req.msg.size(),
@@ -302,7 +302,8 @@ NetworkEngine::requestStep(Sp<Request> sreq)
         err == EHOSTUNREACH ||
         err == EAFNOSUPPORT ||
         err == EPIPE        ||
-        err == EPERM)
+        err == EPERM        ||
+        err == EFAULT)
     {
         node.setExpired();
         if (not node.id)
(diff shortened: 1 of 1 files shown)


src/network_engine.cpp
```diff
@@ -666,9 +667,6 @@ NetworkEngine::send(const char *buf, size_t len, int flags, const SockAddr& addr
     if (sendto(s, buf, len, flags, addr.get(), addr.getLength()) == -1) {
         int err = errno;
         DHT_LOG.e("Can't send message to %s: %s", addr.toString().c_str(), strerror(err));
-        if (err == EPIPE) {
-            throw SocketException(EPIPE);
```
aberaud: EPIPE on UDP sockets is encountered on iOS and macOS devices when getting back from sleep. In this case the socket needs to be recreated. Since the socket is opened and provided by upper layers (DhtRunner) the error has to be propagated back to the DhtRunner,

src/network_engine.cpp
```diff
@@ -290,9 +290,9 @@ NetworkEngine::requestStep(Sp<Request> sreq)
         node.setExpired();
         if (not node.id)
             requests.erase(req.tid);
+        if (req.on_expired)
+            req.on_expired(req, false);
         return;
-    } else if (req.attempt_count == 1) {
-        req.on_expired(req, false);
```
aberaud: Calling on_expired with false is an intended optimisation to allow searches to flag a node as suspicious after the first request timeout, to allow completing the search without waiting for the node to expire. If the node replies before search completion, the flag is cleared and the answer is processed normally. If the node expires, all of its pending requests also expire, and on_expired is called with true.

src/network_engine.cpp
```diff
@@ -290,9 +290,9 @@ NetworkEngine::requestStep(Sp<Request> sreq)
         node.setExpired();
         if (not node.id)
             requests.erase(req.tid);
+        if (req.on_expired)
+            req.on_expired(req, false);
```
aberaud: At this point on_expired was already called (with true) through node.setExpired() -> req->setExpired().

src/network_engine.cpp
hanyf: /* Time for a request to timeout */
    static constexpr const std::chrono::seconds MAX_RESPONSE_TIME {1};

I think 1 second timeout for sending a request is too small, it is inefficient on a network environment.

src/network_engine.cpp
hanyf: Thanks for your explanation. It’s a great help for me to understand how it is working. Thanks!

src/network_engine.cpp
hanyf: Thank you👍

PR #357, opened 2018-12-11
Crypto: move ops to Value

aberaud:


src/securedht.cpp
```diff
@@ -68,33 +68,19 @@ ValueType
...
+        if (v->isSigned())
+            return v->checkSignature();
         return type.storePolicy(id, v, nid, a);
     };
     type.editPolicy = [this,type](InfoHash id, const Sp<Value>& o, Sp<Value>& n, const InfoHash& nid, const SockAddr& a) {
-        if (!o->isSigned())
+        if (not o->isSigned())
             return type.editPolicy(id, o, n, nid, a);
-        if (o->owner != n->owner) {
+        if (o->owner != n->owner or not n->isSigned()) {
```
AmarOk1412: Always false if only one thread, the return is just before

aberaud: This is the value edition policy, where `o` is the old value and `n` the new value.

Maybe `n` still has a owner field (equal to o->owner) but no signature field. In that case n->isSigned would be false.

src/securedht.cpp
```diff
@@ -68,33 +68,19 @@ ValueType
...
-                v->signatureChecked = true;
-                v->signatureValid = v->owner and v->owner->checkSignature(v->getToSign(), v->signature);
-            }
-            if (!v->signatureValid) {
-                DHT_LOG.WARN("Signature verification failed");
-                return false;
-            }
-        }
+        if (v->isSigned())
+            return v->checkSignature();
```
AmarOk1412: No more log if checkSignature fails "Signature verification failed"

src/securedht.cpp
```diff
@@ -68,33 +68,19 @@ ValueType
...
             DHT_LOG.WARN("Edition forbidden: owner changed.");
             return false;
         }
-        if (!n->signatureChecked) {
-            n->signatureChecked = true;
-            n->signatureValid = o->owner and o->owner->checkSignature(n->getToSign(), n->signature);
-        }
-        if (!n->signatureValid) {
-            DHT_LOG.WARN("Edition forbidden: signature verification failed.");
+        if (not n->checkSignature())
```
AmarOk1412: Here too.

src/value.cpp
```diff
@@ -233,6 +233,41 @@ Value::toJson() const
...
 }
 
+bool
+Value::checkSignature()
+{
+    if (!signatureChecked) {
+        signatureChecked = true;
+        if (isSigned()) {
+            signatureValid = owner and owner->checkSignature(getToSign(), signature);
+        } else {
```
AmarOk1412: Remove this block for homogeneization

AmarOk1412:
Else it's ok for me

PR #366, opened 2019-01-07
proxy_server: return old values when resubscribe

AmarOk1412:
This patch is the first part to fix the presence when the android client is using the proxy.

Like a normal listen, we should be able to retrieve old values.


AmarOk1412:
The push notification is only sent once even if there are multiple values. So the `get` will be done only if there is one or more values. 

"Or the subscribe operation could send current values in the http response like a get operation ?" => we can do that yes.

PR #370, opened 2019-01-16
Sync status

aberaud:


AmarOk1412:
Seems ok for me, But not used, don't know if you want to  merge it as it is.

tekknolagi:
Sorry, what does this change do?

aberaud:
@tekknolagi 
In case a `listen` is already ongoing for a given key, a `get` operation for the same key will use the listen cache to answer the `get` operation instead of performing network requests.
This requires to keep track if a listen op is synched for every remote node (any remote storage node responsible for the key could store new values). For instance if a listen was just started the remote nodes are not yet synched so the cache can't be used to complete the get op. immediately.

aberaud:
Note this was actually merged in master independently from this PR.

PR #372, opened 2019-01-17
proxy: only get old values at first subscribe

AmarOk1412:


src/dht_proxy_client.cpp
```diff
@@ -1077,14 +1083,18 @@ DhtProxyClient::getPushRequest(Json::Value& body) const
...
 {
     // Fill body with
     // {
     //   "key":"device_key",
     // }
     Json::Value body;
     getPushRequest(body);
+    if (!resubscribe) {
+        // This is the first listen, we want to retrieve previous values.
+        body["previous_values"] = "on";
```
aberaud: true

src/dht_proxy_server.cpp
```diff
@@ -429,22 +429,37 @@ DhtProxyServer::subscribe(const std::shared_ptr<restbed::Session>& session)
...
-                                    return true;
-                                }, [s](bool /*ok* */) {
-                                    // Communication is finished
-                                    if (not s->is_closed()) {
-                                        s->close("{}\n");
-                                    }
-                                });
+
+                            if (root.isMember("previous_values") &&
+                                root["previous_values"].asString() == "on") {
```
aberaud: asBool

src/dht_proxy_server.cpp
```diff
@@ -429,22 +429,37 @@ DhtProxyServer::subscribe(const std::shared_ptr<restbed::Session>& session)
...
+                                    [s](bool /*ok* */) {
+                                        // Communication is finished
+                                        if (not s->is_closed()) {
+                                            s->close("{}\n");
+                                        }
+                                    });
+                            } else {
+                                // Communication is finished
+                                if (not s->is_closed()) {
+                                    s->close("{}\n");
```
aberaud: Why not just `s->close();` or `s->close("");` ?

AmarOk1412: Because it will be interpreted as a failure client side and will loop (`reader->parse` will throw, and you will have
```
DHT_LOG.w("Listen closed by the proxy server: %s", e.what());
                        state->ok = false;
```)

PR #373, opened 2019-01-22
proxy_client: support prototype in serverHost

AmarOk1412:


src/dht_proxy_client.cpp
```diff
@@ -80,6 +80,8 @@ DhtProxyClient::DhtProxyClient() {}
 DhtProxyClient::DhtProxyClient(std::function<void()> signal, const std::string& serverHost, const std::string& pushClientId)
 : serverHost_(serverHost), pushClientId_(pushClientId), loopSignal_(signal)
 {
+    if (serverHost_.find("://") == std::string::npos)
+        serverHost_ = proxy::HTTP_PROTO + serverHost_;
```
aberaud: maybe we should add the sheme only if not already present (and let the request fail if the sheme is not supported by restbed) ?

PR #378, opened 2019-02-18
dhtrunner: avoid startNetwork failure if only one binding is failing

AmarOk1412:


src/dhtrunner.cpp
```diff
@@ -531,17 +534,35 @@ DhtRunner::startNetwork(const SockAddr sin4, const SockAddr sin6)
...
+        try {
+            s6 = bindSocket(sin6, bound6);
+        } catch (const DhtException& e) {
+           inet6BindException = e;
+        }
+    }
 #endif
 
+    if (s4 == -1 && s6 == -1) {
+        if (s4 == -1)
```
aberaud: always true

src/dhtrunner.cpp
```diff
@@ -531,17 +534,35 @@ DhtRunner::startNetwork(const SockAddr sin4, const SockAddr sin6)
...
+            throw inetBindException;
+        if (s6 == -1)
```
aberaud: dead code (inetBindException always thrown)

src/dhtrunner.cpp
```diff
@@ -531,17 +534,35 @@ DhtRunner::startNetwork(const SockAddr sin4, const SockAddr sin6)
...
+    DhtException inetBindException, inet6BindException;
 
     bound4 = {};
-    if (sin4)
-        s4 = bindSocket(sin4, bound4);
+    if (sin4) {
+        try {
+            s4 = bindSocket(sin4, bound4);
+        } catch (const DhtException& e) {
+           inetBindException = e;
```
aberaud: just print the socket-specific error there and throw a generic "can't bind socket" exception later ?

aberaud:
merged in master

PR #381, opened 2019-02-25
Persist state

aberaud:


AmarOk1412:
(I'll review this tonight)

tools/tools_common.h
```diff
@@ -154,6 +156,9 @@ parseArgs(int argc, char **argv) {
         case 'D':
             params.devicekey = optarg;
             break;
+        case 'f':
```
AmarOk1412: Missing "f" in getopt_long. (--persist is recognized, -f is not)

src/dht.cpp
```diff
@@ -2424,4 +2431,52 @@ Dht::onAnnounceDone(const Sp<Node>& node, net::RequestAnswer& answer, Sp<Search>
...
+{
+    std::ofstream file(path);
+    msgpack::pack(file, exportNodes());
+    msgpack::pack(file, exportValues());
+}
+
+void
+Dht::loadState(const std::string& path)
+{
+    DHT_LOG.d("Importing state from %s", path.c_str());
```
AmarOk1412: Note: these logs are not shown when using dhtnode.cpp, in fact loggers are enabled after `dht->run()` so loadState is already called.

aberaud: Indeed.. I plan to add the logger as part of the dht constructor or config (might be in another patch)

AmarOk1412: yup not related to this PR (that's why I just add a Note). I've just noticed this weird thing.

tools/tools_common.h
aberaud: done

PR #386, opened 2019-03-17
Network: improve parsing performance

aberaud:


aberaud:
On my computer, benchmarks show 100% (doubled) packet parsing performance improvement (= packet parsing time reduced by 50%, 30% if including msgpack parsing).

PR #387, opened 2019-03-28
 crypto: add certificate request support

aberaud:


src/crypto.cpp
```diff
@@ -572,6 +572,153 @@ PublicKey::getLongId() const
...
+        return NameType::URI;
+    case GNUTLS_SAN_IPADDRESS:
+        return NameType::IP;
+    default:
+        return NameType::UNKNOWN;
+    }
+}
+
+static gnutls_x509_subject_alt_name_t
+typeFromName(NameType type)
```
AmarOk1412: `typeFromGnuTLS` <=> `GnuTLSFromName` maybe here?

The previous function is the invert operation, should be reflected into the API

src/crypto.cpp
```diff
@@ -572,6 +572,153 @@ PublicKey::getLongId() const
...
+CertificateRequest::sign(const PrivateKey& key, const std::string& password)
+{
+    gnutls_x509_crq_set_version(request, 1);
+    if (not password.empty())
+        gnutls_x509_crq_set_challenge_password(request, password.c_str());
+
+    if (auto err = gnutls_x509_crq_set_key(request,  key.x509_key))
+        throw CryptoException(std::string("Can't set certificate request key: ") + gnutls_strerror(err));
+        
+    if (auto err = gnutls_x509_crq_sign(request, key.x509_key))
```
AmarOk1412: Seems deprecated. `gnutls_x509_crq_privkey_sign` instead.

Source https://manpages.debian.org/stretch/gnutls-doc/gnutls_x509_crq_sign.3.en.html

src/crypto.cpp
aberaud: done

src/crypto.cpp
aberaud: done

AmarOk1412:
Ok for me!

PR #390, opened 2019-04-05
peerdiscovery: add peerdiscovery class

MingruiZhangW:
PeerDiscovery Functions are mostly used in the dhtrunner,
once node.run() is called, peerDiscovery will start with one (two, actually with ipv4, ipv6) listener and one sender to send and receive node id and node port data and insert the node into their corresponding routing table. These threads will be terminated once the node.join() is called.

New Pull request to avoid the original mess


MingruiZhangW:
Seems like the dhtrunner test may still fail sometimes,
but 4 out of 5

aberaud:
merged in master

PR #396, opened 2019-04-30
thread pool: add executor

aberaud:


AmarOk1412:
Works perfectly here :)

PR #422, opened 2019-08-20
tester: add https support to proxy_node

binarytrails:
Tested in chrome by allowing it: [chrome://flags/#allow-insecure-localhost]()


binarytrails:
Tested in chrome by allowing it: [chrome://flags/#allow-insecure-localhost`]()

PR #423, opened 2019-08-22
autotools: fix dhtproxy build

binarytrails:


configure.ac
```diff
@@ -142,25 +142,26 @@ AS_IF([test "x$have_jsoncpp" = "xyes"], [
 ])
 
 AC_ARG_WITH([openssl], AS_HELP_STRING([--with-openssl], [Build with OpenSSL support]))
-AS_IF([test "x$with_openssl" != "xno"],
+AS_IF([test "x$with_openssl" = "xyes"],
```
aberaud: Should default to use openssl except if --without-openssl is passed.
This would build with openssl only if --with-openssl is used

configure.ac
```diff
@@ -142,25 +142,26 @@ AS_IF([test "x$have_jsoncpp" = "xyes"], [
...
       [PKG_CHECK_MODULES([OpenSSL], [openssl >= 1.1], [have_openssl=yes], [have_openssl=no])],
       [have_openssl=no])
 AS_IF([test "x$have_openssl" = "xyes"], [
+    LDFLAGS="${LDFLAGS} -lssl"
```
aberaud: PKG_CHECK_MODULES should set macros OpenSSL_CFLAGS, OpenSSL_LIBS and OpenSSL_LDFLAGS (see for other dependencies like Argon2)

configure.ac
```diff
@@ -142,25 +142,26 @@ AS_IF([test "x$have_jsoncpp" = "xyes"], [
...
 ], [
     AC_MSG_NOTICE([Not using OpenSSL])
 ])
 
 AC_ARG_WITH([http_parser_fork], AS_HELP_STRING([--with-http-parser-fork], [Build with http_parser fork to support old API]))
-AS_IF([test "x$with_http_parser_fork" != "xno"],[
+AS_IF([test "x$with_http_parser_fork" = "xyes"],[
     AC_MSG_NOTICE([Using http_parser fork])
-    CPPFLAGS+=" -DOPENDHT_PROXY_HTTP_PARSER_FORK"
+    CPPFLAGS+=" -DOPENDHT_PROXY_HTTP_PARSER_FORK=1"
```
aberaud: should not be necessary

configure.ac
```diff
@@ -201,6 +202,7 @@ AM_COND_IF([ENABLE_TOOLS], [
 
 AM_COND_IF(ENABLE_PROXY_SERVER, [CPPFLAGS+=" -DOPENDHT_PROXY_SERVER"], [])
 AM_COND_IF(ENABLE_PROXY_CLIENT, [CPPFLAGS+=" -DOPENDHT_PROXY_CLIENT"], [])
+AM_COND_IF(PROXY_CLIENT_OR_SERVER, [CPPFLAGS+=" -DPROXY_CLIENT_OR_SERVER"], [])
```
aberaud: never used in source code

binarytrails: @aberaud Oh I thought I needed it here [src/Makefile.am#L75](https://github.com/binarytrails/opendht/blob/proxy_autotools/src/Makefile.am#L75) to enter this conditional

configure.ac
binarytrails: Done at ad695035aa25ff725c51447a3aa5fb5241e9cfe2

configure.ac
binarytrails: Done at ad695035aa25ff725c51447a3aa5fb5241e9cfe2

configure.ac
binarytrails: Done at ad695035aa25ff725c51447a3aa5fb5241e9cfe2

configure.ac
binarytrails: Fixed at 3dcca38221cd02784a129b4782fcc8e90d3daeac

PR #432, opened 2019-09-05
WIP dhtproxy: add two-ways tls & expose verify cb in api

binarytrails:
Depends-On: #425


include/opendht/dht_proxy_server.h
```diff
@@ -74,7 +74,7 @@ class OPENDHT_PUBLIC DhtProxyServer
     DhtProxyServer(
        dht::crypto::Identity identity,
        std::shared_ptr<DhtRunner> dht, in_port_t port = 8000, const std::string& pushServer = "",
-       std::shared_ptr<dht::Logger> logger = {});
+       const std::string& client_certificate = "", std::shared_ptr<dht::Logger> logger = {});
```
aberaud: should be able to provide in-memory certificate (from shared_ptr<crypto::Certificate>)

binarytrails: @aberaud we could use the in-memory feature to add a client certificate to the ssl strore [1] however, this would require design changes to update them dynamically on the server instance. An interesting thread explaining the underlying issue is here [2] and the solution to update them dynamically would be at [1]. To recall this feature is to be able to validate multiple clients with their certificates at the store.
[1] http://think-async.com/Asio/asio-1.12.2/doc/asio/reference/ssl__context/add_certificate_authority/overload2.html
[2] http://boost.2283326.n4.nabble.com/ASIO-ssl-load-verify-file-td2577261.html

PR #436, opened 2019-09-12
http: handle expect 100-continue on post

binarytrails:
> A 100-continue expectation informs recipients that the client is about to send a (presumably large) message body in this request and wishes to receive a 100 (Continue) interim response if the request-line and header fields are not sufficient to cause an immediate success, redirect, or error response.
https://tools.ietf.org/html/rfc7231#section-5.1.1

Adds feature: ```request->set_header_field(restinio::http_field_t::expect, "100-continue");```

### DIFF
--- a/src/http.cpp
+++ b/src/http.cpp
@@ -43,8 +43,10 @@ Url::Url(const std::string& url): url(url)
     const size_t proto_end = url.find("://");
     if (proto_end != std::string::npos){
         addr_begin = proto_end + 3;
-        if (url.substr(0, proto_end) == "https")
+        if (url.substr(0, proto_end) == "https"){
             protocol = "https";
+            service = protocol;
+        }
     }
     // host and service
     size_t addr_size = url.substr(addr_begin).find("/");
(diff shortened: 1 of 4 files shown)


src/http.cpp
```diff
@@ -546,13 +556,11 @@ Request::build()
...
     // body & content-length
-    if (!body_.empty()){
+    if (!body_.empty())
         request << "Content-Length: " << body_.size() << "\r\n\r\n";
-        request << body_;
-    }
-
     // last delim
-    request << "\r\n";
+    if (append_body)
```
AmarOk1412: body_ should be non empty.

binarytrails: @AmarOk1412 in a case of 100-continue yes! its the purpose of such optimization is send the header only before continuing with a body, thats why I set the variable append_body in such a header is present! line 538: ```if (header.first == restinio::http_field_t::expect and header.second == "100-continue")```

AmarOk1412: I know. What I mean is to add the append_body only if body_.empty() is also false; Like

if (append_body && !body.empty())

binarytrails: but then, I will have to trim the raw request that will be written on a socket during a send(), thus, I just build() one without a body and send the body separately after in the sequence if such option is set!

PR #440, opened 2019-09-18
proxyserver: default push to localhost if empty host

binarytrails:


src/dht_proxy_server.cpp
```diff
@@ -222,6 +222,14 @@ DhtProxyServer::DhtProxyServer(
     jsonBuilder_["commentStyle"] = "None";
     jsonBuilder_["indentation"] = "";
 
+    // no host delim, assume port only
+    if (pushServer.find(":") == std::string::npos)
+        pushServer_ =  "localhost:" + pushServer_;
+    // define http request destination for push notifications
+    pushHostPort_ = splitPort(pushServer_);
+    if (logger_)
+        logger_->d("Using push server for notifications: %s:%s", pushHostPort_.first.c_str(),
```
aberaud: only log if pushServer_ not empty

PR #441, opened 2019-09-18
msvc: fix compilation with VS 2019 16.0

atraczyk:
- <stdexcept> is no longer included in <vector> as of vs 2019


include/opendht/sockaddr.h
```diff
@@ -40,6 +40,9 @@ typedef uint16_t in_port_t;
 #include <string>
 #include <memory>
 #include <vector>
+#if (_MSC_VER >= 1920) // VS 2019 16.0 no longer includes <stdexcept> in <vector>
```
aberaud: macro check not needed

PR #446, opened 2019-09-28
http: implement chunked transfer-encoding

binarytrails:


src/http.cpp
```diff
@@ -922,7 +924,17 @@ Request::handle_response_header(const asio::error_code& ec)
             std::bind(&Request::handle_response_body, this, std::placeholders::_1, std::placeholders::_2));
     }
     else if (connection_type_ == restinio::http_connection_header_t::close)
+    {
         terminate(asio::error::eof);
+    }
+    else if (response_.headers[HTTP_HEADER_TRANSFER_ENCODING] == HTTP_HEADER_TRANSFER_ENCODING_CHUNKED)
```
aberaud: `operator[]` would create `headers` item if it doesn't exist (same for line 921 https://github.com/savoirfairelinux/opendht/pull/446/files#diff-6db9a270f82169aa49daea29e8c5be13R921 ).
Prefer `find` like line 903  https://github.com/savoirfairelinux/opendht/pull/446/files#diff-6db9a270f82169aa49daea29e8c5be13R903

src/http.cpp
```diff
@@ -959,7 +971,13 @@ Request::handle_response_body(const asio::error_code& ec, const size_t bytes)
     // read and parse the chunked encoding fragment
     else {
         auto body = conn_->read_until(JSON_VALUE_DELIM[0]) + '\n';
```
aberaud: JSON_VALUE_DELIM has wrong name (not specific to json)

src/http.cpp
```diff
@@ -977,7 +995,23 @@ Request::handle_response_body(const asio::error_code& ec, const size_t bytes)
             std::bind(&Request::handle_response_body, this, std::placeholders::_1, std::placeholders::_2));
     }
     else if (connection_type_ == restinio::http_connection_header_t::close)
+    {
         terminate(asio::error::eof);
+    }
+    else if (response_.headers[HTTP_HEADER_TRANSFER_ENCODING] == HTTP_HEADER_TRANSFER_ENCODING_CHUNKED)
```
aberaud: operator[] would create headers item if it doesn't exist (also for line 993)

src/http.cpp
```diff
@@ -977,7 +995,23 @@ Request::handle_response_body(const asio::error_code& ec, const size_t bytes)
...
+        std::istream is(&conn_->data());
+        std::string chunk_size;
+        std::getline(is, chunk_size);
+        if (chunk_size.size() == 0){
+            std::replace(response_.body.begin(), response_.body.end(), '\n', '\0');
+            parse_request(response_.body);
+            terminate(asio::error::eof);
+        }
+        else
+            conn_->async_read_until(JSON_VALUE_DELIM,
```
aberaud: should use `async_read` with `chunk_size` ?

src/http.cpp
```diff
@@ -977,7 +995,23 @@ Request::handle_response_body(const asio::error_code& ec, const size_t bytes)
...
+    {
         terminate(asio::error::eof);
+    }
+    else if (response_.headers[HTTP_HEADER_TRANSFER_ENCODING] == HTTP_HEADER_TRANSFER_ENCODING_CHUNKED)
+    {
+        std::istream is(&conn_->data());
+        std::string chunk_size;
+        std::getline(is, chunk_size);
+        if (chunk_size.size() == 0){
+            std::replace(response_.body.begin(), response_.body.end(), '\n', '\0');
```
aberaud: What does this do ?

binarytrails: it ensures that if there is no content-length following a body read aka the line is empty, we will not continue reading and consider that we received all of the chunks. The server is normally suppose to send 0\r\n which our tomcat of jams did but over the weekend only the 8080 was available and if you perform a curl on the functionality of chunked encoding while getting the devices, it does not do that and simply finishes by not writing nothing.

src/http.cpp
```diff
@@ -916,13 +922,25 @@ Request::handle_response_header(const asio::error_code& ec)
...
-        conn_->async_read_until(JSON_VALUE_DELIM,
+        conn_->async_read_until(BODY_VALUE_DELIM,
             std::bind(&Request::handle_response_body, this, std::placeholders::_1, std::placeholders::_2));
     }
     else if (connection_type_ == restinio::http_connection_header_t::close)
+    {
         terminate(asio::error::eof);
+    }
+    else if (transfer_encoding_it != response_.headers.end() and
+             response_.headers[HTTP_HEADER_TRANSFER_ENCODING] == HTTP_HEADER_TRANSFER_ENCODING_CHUNKED)
```
aberaud: use iterator to access map value:
*transfer_encoding_it == HTTP_HEADER_TRANSFER_ENCODING_CHUNKED

PR #447, opened 2019-09-29
dhtproxy: test put/get mtu with 40k chars

binarytrails:


tests/dhtproxytester.cpp
```diff
@@ -207,4 +207,50 @@ DhtProxyTester::testResubscribeGetValues() {
...
+    bool done_get = false;
+
+    // Act
+    dht::Value val {mtu};
+    nodeClient->put(key, std::move(val), [&](bool ok) {
+        std::lock_guard<std::mutex> lk(cv_m);
+        done_put = ok;
+        cv.notify_all();
+    });
+    cv.wait_for(lk, std::chrono::seconds(10), [&]{ return done_put; });
```
aberaud: CPPUNIT_ASSERT(cv.wait_for(...))

This way the test fails if the timeout is reached and done_put is false

tests/dhtproxytester.cpp
```diff
@@ -207,4 +207,50 @@ DhtProxyTester::testResubscribeGetValues() {
...
+
+    nodeClient->get(key, [&](const std::vector<std::shared_ptr<dht::Value>>& vals){
+        values = vals;
+        return true;
+    },[&](bool ok){
+        done_get = ok;
```
aberaud: lock mutex while changing value

tests/dhtproxytester.cpp
```diff
@@ -207,4 +207,50 @@ DhtProxyTester::testResubscribeGetValues() {
...
+        cv.notify_all();
+    });
+    cv.wait_for(lk, std::chrono::seconds(10), [&]{ return done_get; });
```
aberaud: CPPUNIT_ASSERT(cv.wait_for(...))

PR #449, opened 2019-10-04
WIP http: implement redirects

binarytrails:
- ~~301 moved permanently~~ (tested with google.com -> www.google.com)
- 302 found
- 307 temporary redirect
- 308 permanent redirect


aberaud:
implemented in another way, to allow redirecting from http to https, or redirect to another server

PR #450, opened 2019-10-07
http: close connection if server requests

binarytrails:
- tested with http://httpstat.us/400 etc
- client test sample: [http_errors.cpp](https://github.com/binarytrails/various/blob/master/cpp/opendht/http_errors.cpp
)
- test all 4xx 5xx codes with ```for code in $(seq 400 500); do ./http_errors.out $code; done```

### DIFF
--- a/src/http.cpp
+++ b/src/http.cpp
@@ -31,6 +31,7 @@ namespace http {
 
 constexpr char HTTP_HEADER_CONNECTION[] = "Connection";
 constexpr char HTTP_HEADER_CONNECTION_KEEP_ALIVE[] = "keep-alive";
+constexpr char HTTP_HEADER_CONNECTION_CLOSE[] = "close";
 constexpr char HTTP_HEADER_CONTENT_LENGTH[] = "Content-Length";
 constexpr char HTTP_HEADER_CONTENT_TYPE[] = "Content-Type";
 constexpr char HTTP_HEADER_CONTENT_TYPE_JSON[] = "application/json";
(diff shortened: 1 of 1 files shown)


src/http.cpp
```diff
@@ -921,6 +921,10 @@ Request::handle_response_header(const asio::error_code& ec)
                 std::bind(&Request::handle_response_body, this, std::placeholders::_1, std::placeholders::_2));
         }
     }
+    else if (response_.status_code >= 400 and response_.status_code <= 500)
```
aberaud: response with no body should be properly handled regardless of the HTTP response code

src/http.cpp
```diff
@@ -998,6 +1002,10 @@ Request::handle_response_body(const asio::error_code& ec, const size_t bytes)
         conn_->async_read(content_length - response_.body.size(),
             std::bind(&Request::handle_response_body, this, std::placeholders::_1, std::placeholders::_2));
     }
+    else if (response_.status_code >= 400 and response_.status_code <= 500)
+    {
+        terminate(asio::error::operation_aborted);
```
aberaud: response with body should be properly handled regardless of the HTTP response code

src/http.cpp
binarytrails: @aberaud I think the empty body aka just headers are well handled in the lines before as tested with [http_errors.cpp](https://github.com/savoirfairelinux/opendht/pull/450) !

src/http.cpp
binarytrails: @aberaud I think the body cases are well handled in the lines before as tested with [http_errors.cpp](https://github.com/binarytrails/various/blob/master/cpp/opendht/http_errors.cpp) because this condition is in else if after all possible body reads

PR #458, opened 2019-10-25
Link to certificate C-bindings

AmarOk1412:


rust/src/crypto.rs
```diff
@@ -111,4 +114,58 @@ impl Drop for PrivateKey {
...
+impl Drop for DhtCertificate {
+    fn drop(&mut self) {
+        unsafe {
+            dht_certificate_delete(&mut *self)
+        }
+    }
+}
+
+impl DhtIdentity {
+    pub fn generate(common_name: &str, ca: Box<DhtIdentity>) -> DhtIdentity {
```
aberaud: `ca` can be null

rust/src/crypto.rs
```diff
@@ -111,4 +114,58 @@ impl Drop for PrivateKey {
             dht_privatekey_delete(&mut *self)
         }
     }
+}
+
+impl DhtCertificate {
+    pub fn import(file: &str) -> io::Result<Box<DhtCertificate>> {
```
aberaud: would be nice to be able to load from memory

rust/src/crypto.rs
AmarOk1412: Done

rust/src/crypto.rs
AmarOk1412: Done

rust/src/crypto.rs
```diff
@@ -111,4 +115,68 @@ impl Drop for PrivateKey {
...
+        let mut f = File::open(file)?;
+        let mut buffer = Vec::new();
+        f.read_to_end(&mut buffer)?;
+        Ok(DhtCertificate::from_bytes(&buffer))
+    }
+
+    pub fn from_bytes(buffer: &Vec<u8>) -> Box<DhtCertificate> {
```
aberaud: allow to import from slice ?

rust/src/infohash.rs
```diff
@@ -47,6 +47,14 @@ impl InfoHash {
         h
     }
 
+    pub fn from_bytes(data: &Vec<u8>) -> InfoHash {
```
aberaud: allow to import from slice ?

PR #472, opened 2019-12-13
http_parser: install in /usr not /usr/local

AmarOk1412:


aberaud:
merged in another patch

PR #476, opened 2020-01-06
Peer discovery

257:
implement connectivity change in peer discovery

this makes peer discovery pub/sub on connectivity change, it aims to
make peer discovery more energy efficient.

TODO:
in its current implementation however, peer_discovery operates in the
dark; it doesn't know routes available to it (v4 vs v6); nor does it
know of the state of the link(s) (UP vs. DOWN) and therefore it merely
tries to, on connectivity change, (re)pub/sub on previous ip version
routes.

IDEA:
open an interface to kernel so that peer discovery, and perhaps others,
can learn about the routes available to them.

### DIFF
--- a/src/peer_discovery.cpp
+++ b/src/peer_discovery.cpp
@@ -1,5 +1,5 @@
 /*
- *  Copyright (C) 2014-2019 Savoir-faire Linux Inc.
+ *  Copyright (C) 2014-2020 Savoir-faire Linux Inc.
  *  Author(s) : Mingrui Zhang <mingrui.zhang@savoirfairelinux.com>
  *              Vsevolod Ivanov <vsevolod.ivanov@savoirfairelinux.com>
  *              Adrien Béraud <adrien.beraud@savoirfairelinux.com>
--- a/src/peer_discovery.cpp
+++ b/src/peer_discovery.cpp
@@ -46,6 +46,8 @@ public:
     bool stopDiscovery(const std::string &type);
     bool stopPublish(const std::string &type);
 
+    void connectivityChanged();
+
 private:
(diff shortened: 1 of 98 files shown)


src/peer_discovery.cpp
```diff
@@ -247,6 +245,37 @@ PeerDiscovery::DomainPeerDiscovery::reloadMessages()
     }
 }
 
+void
+PeerDiscovery::DomainPeerDiscovery::reDiscover()
+{
+	asio::error_code ec;
+
+	sockFd_.set_option(asio::ip::multicast::leave_group(sockAddrSend_.address()), ec);
+	if (ec)
```
AmarOk1412: if (ec and logger_)

indentation

src/peer_discovery.cpp
```diff
@@ -247,6 +245,37 @@ PeerDiscovery::DomainPeerDiscovery::reloadMessages()
...
+			logger_->w("can't multicast on %s: %s",
+					sockAddrSend_.address().to_string(),
+					ec.message().c_str());
+}
+
+void
+PeerDiscovery::DomainPeerDiscovery::rePublish()
+{
+   publish();
+}
```
AmarOk1412: Just use publish if nothing is done

src/peer_discovery.cpp
257: i followed line 163+164's logic. but yes ```if (ec and logger_)``` will catch the exception also.
anyway, do you want this to be changed also?
from
```
if (ec and ec != asio::error::operation_aborted)
    if (logger)
```
to
```
if (ec and (ec != asio::error::operation_aborted) and logger)
```

PR #477, opened 2020-01-07
peer discovery

257:
clean up

### DIFF
--- a/src/peer_discovery.cpp
+++ b/src/peer_discovery.cpp
@@ -46,6 +46,8 @@ public:
     bool stopDiscovery(const std::string &type);
     bool stopPublish(const std::string &type);
 
+    void connectivityChanged();
+
 private:
     Sp<Logger> logger_;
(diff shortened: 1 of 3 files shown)


src/peer_discovery.cpp
```diff
@@ -348,4 +369,13 @@ PeerDiscovery::stopPublish(sa_family_t domain, const std::string& type)
     return false;
 }
 
+void
+PeerDiscovery::connectivityChanged()
+{
+    if (peerDiscovery4_)
+        peerDiscovery4_->connectivityChanged();
```
aberaud: can throw

src/peer_discovery.cpp
```diff
@@ -348,4 +369,13 @@ PeerDiscovery::stopPublish(sa_family_t domain, const std::string& type)
...
+    if (peerDiscovery6_)
+        peerDiscovery6_->connectivityChanged();
```
aberaud: can throw

PR #481, opened 2020-02-05
ConnectivityStatus

257:
this will help getting more granular reporting of connectivity status
from kernel.  it is linux/android specific for now with possible
future support for bsd/darwin.

this is very much a prototype; cleaner interface and code will follow.

it adds libnl (ships with android) dependency which is really not
required, although expedited prototyping.


cmake/FindNL.cmake
```diff
@@ -0,0 +1,144 @@
...
+	set(HAVE_LIBNL 1)
+else()
+	set(NL_LIBRARIES )
+	set(NL_INCLUDE_DIRS)
+endif()
+
+MARK_AS_ADVANCED( NL_LIBRARIES NL_INCLUDE_DIRS )
+
+# MESSAGE(STATUS "NL_LIBRARIES ${NL_LIBRARIES}")
+# MESSAGE(STATUS "NL_INCLUDE_DIRS ${NL_INCLUDE_DIRS}")
```
AmarOk1412: Debug message? Not necessary anymore

include/opendht/connstat.h
```diff
@@ -0,0 +1,98 @@
...
+struct nl_sock;
+struct nl_msg;
+
+namespace dht {
+namespace net {
+
+class OPENDHT_PUBLIC ConnectivityStatus
+{
+public:
+    enum topic {
```
AmarOk1412: enum class Topic

include/opendht/connstat.h
```diff
@@ -0,0 +1,98 @@
...
+
+    using connstat_event_cb = std::function<void (unsigned int)>;
+
+    void setTopicListener    (connstat_event_cb ucb, unsigned int topic);
+    void removeTopicListener (unsigned int topic);
+
+    int  nl_event_cb         (struct nl_msg*);
+
+private:
+    enum status { SUCCESS, FAILURE };
```
AmarOk1412: bool

src/connstat.cpp
```diff
@@ -0,0 +1,187 @@
...
+
+/* forward declarations */
+static inline uint32_t nl_mgrp(uint32_t group);
+static int nlcb(struct nl_msg*, void*);
+
+ConnectivityStatus::ConnectivityStatus()
+    : nlsk(nlsk_init())
+    , logger_(new dht::Logger)
+{
+    if (nlsk_setup(nlsk.get()) == FAILURE)
```
AmarOk1412: if (not nlsk_setup(nlsk.get()))

include/opendht/connstat.h
```diff
@@ -0,0 +1,98 @@
...
+
+    connstat_event_cb topic_cbs[CONNSTAT_TOPICS_MAX] = {nullptr};
```
AmarOk1412: std::array is here

AmarOk1412: {nullptr} => {}

src/connstat.cpp
```diff
@@ -0,0 +1,187 @@
...
+    topic_cbs[topic] = ucb;
+}
+
+void
+ConnectivityStatus::removeTopicListener(unsigned int topic)
+{
+    switch (topic) {
+    case NEWADDR:
+        nl_socket_drop_memberships(nlsk.get(), RTNLGRP_IPV6_IFADDR, RTNLGRP_NONE);
+        topic_cbs[topic] = nullptr;
```
AmarOk1412: nullptr => {}

src/connstat.cpp
```diff
@@ -0,0 +1,187 @@
...
+}
+
+void
+ConnectivityStatus::executer(unsigned int topic)
+{
+    if (*(topic_cbs+topic))
+        (*(topic_cbs+topic))(topic);
+}
+
+int
```
AmarOk1412: bool

AmarOk1412: unless if you will return a enum status, (if so just return a status)

src/connstat.cpp
```diff
@@ -0,0 +1,187 @@
...
+     * Notifications do not use sequence numbers, disable sequence number
+     * checking.
+     */
+    nl_socket_disable_seq_check(nlsk);
+
+    /*
+     * Define a callback function, which will be called for each notification
+     * received
+     */
+    nl_socket_modify_cb(nlsk, NL_CB_VALID, NL_CB_CUSTOM, nlcb, (void *)this);
```
AmarOk1412: Just for info: https://en.cppreference.com/w/cpp/language/explicit_cast

AmarOk1412: Same for other cast

src/connstat.cpp
```diff
@@ -0,0 +1,187 @@
...
+    return nl_mgrp(topic);
+}
+
+
+
+static inline uint32_t nl_mgrp(uint32_t group)
+{
+    if (group > 31 ) {
+        fprintf(stderr, "Use setsockopt for this group %d\n", group);
+        exit(-1);
```
AmarOk1412: cpp code, no C :) (replace exit, fprintf)

cmake/FindNL.cmake
257: done

src/connstat.cpp
257: removed

src/connstat.cpp
257: good read, thanks.

src/connstat.cpp
257: done

include/opendht/connstat.h
257: done

configure.ac
```diff
@@ -18,8 +18,9 @@ AC_SUBST(OPENDHT_PATCH_VERSION, opendht_patch_version)
 
 AC_ARG_ENABLE([debug], AS_HELP_STRING([--enable-debug], [Build in debug mode, adds stricter warnings, disables optimization]))
 AS_IF([test "x$enable_debug" = "xyes"],
-      [CXXFLAGS="${CXXFLAGS} -g -Wno-return-type -Wall -Wextra -Wnon-virtual-dtor -O0 -pedantic-errors"],
-      [CXXFLAGS="${CXXFLAGS} -O3 -pedantic-errors"])
+      [
+       CXXFLAGS="${CXXFLAGS} -g -Wno-return-type -Wall -Wextra -Wnon-virtual-dtor -O0 -pedantic-errors -pedantic -ggdb -fsanitize=address"],
+       [CXXFLAGS="${CXXFLAGS} -O3 -pedantic-errors"])
```
aberaud: cleanup debug stuff

include/opendht/connstat.h
```diff
@@ -0,0 +1,97 @@
...
+struct nl_sock;
+struct nl_msg;
+
+namespace dht {
+namespace net {
+
+class OPENDHT_PUBLIC ConnectivityStatus
+{
+public:
+    enum class topic : const long unsigned int {
```
aberaud: `Topic` (Use CapitalizedCamelCase for type names)
`const long unsigned int` should not be required (if it is required, just use `unsigned long`)

include/opendht/connstat.h
```diff
@@ -0,0 +1,97 @@
...
+        TOPICS_DEL,
+        DELLINK,
+        DELROUTE4,
+        DELROUTE6,
+        DELROUTE,
+        DELADDR4,
+        DELADDR6,
+        DELADDR,
+        DELNEIGH,
+        TOPICS_MAX
```
aberaud: TOPICS_MAX should not be needed

include/opendht/connstat.h
```diff
@@ -0,0 +1,97 @@
...
+    };
+
+    ConnectivityStatus();
+
+    using connstat_event_cb = std::function<void (long unsigned int)>;
```
aberaud: Use CapitalizedCamelCase for type names. Use Topic for the callback type: 

`using ConnectionEventCb = std::function<void(Topic)>;
`

include/opendht/connstat.h
```diff
@@ -0,0 +1,97 @@
...
+
+    void setTopicListener    (connstat_event_cb ucb, topic);
+    void removeTopicListener (topic);
+
+    bool nl_event_cb         (struct nl_msg*);
+
+private:
+    bool SUCCESS, FAILURE;
```
aberaud: ???

include/opendht/connstat.h
```diff
@@ -0,0 +1,97 @@
...
+
+    static const std::size_t topics_max = static_cast<std::size_t>(topic::TOPICS_MAX);
+    std::array<connstat_event_cb, topics_max> topic_cbs = {};
```
aberaud: use `std::map<Topic, ConnectivityEventCb>`

src/connstat.cpp
```diff
@@ -0,0 +1,169 @@
...
+    /*
+     * Start receiving messages. The function nl_recvmsgs_default() will block
+     * until one or more netlink messages (notification) are received which
+     * will be passed on to nlcb().
+     * 0 on success or a negative error code from nl_recv():
+     * Number of bytes read,
+     * 0 on EOF,
+     * 0 on no data event (non-blocking mode), or
+     * a negative error code.
+     */
```
aberaud: remove comment

src/connstat.cpp
```diff
@@ -0,0 +1,169 @@
...
+
+ConnectivityStatus::ConnectivityStatus()
+    : nlsk(nlsk_init())
+    , logger_(new dht::Logger)
+{
+    if (not nlsk_setup(nlsk.get()))
+        throw std::runtime_error("nlsk_setup() failed\n");
+
+    thrd_ = std::thread([this] () { nl_event_loop_thrd(nlsk.get()); });
+    thrd_.detach();
```
aberaud: never detach thread, especially since you capture this (would cause use-after-free).

src/connstat.cpp
```diff
@@ -0,0 +1,169 @@
...
+            return get_neigh_state(msg);
+            break;
+        default:
+            break;
+    }
+
+    return SUCCESS;
+}
+
+static int nlcb(struct nl_msg* msg, void* arg)
```
aberaud: no need to split declaration/definition (could even be inlined as a lambda in call to  nl_socket_modify_cb)

src/connstat.cpp
```diff
@@ -0,0 +1,169 @@
...
+        break;
+    default:
+        break;
+    }
+}
+
+void
+ConnectivityStatus::executer(topic topic)
+{
+    long unsigned int t = static_cast<unsigned int>(topic);
```
aberaud: cast from/to integer not needed

include/opendht/connstat.h
257: moved to ```std::map```

src/connstat.cpp
257: without it i get:
_terminate called without an active exception_

src/connstat.cpp
```diff
@@ -0,0 +1,151 @@
...
+            executer(Event::ADDR);
+            executer(Event::DELADDR);
+            break;
+        case RTM_NEWNEIGH:
+        case RTM_DELNEIGH:
+        case RTM_NEWNEIGHTBL:
+            get_neigh_state(msg);
+            break;
+        default:
+            status = 1;
```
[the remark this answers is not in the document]
257: switched to using ```nl_cb_action```

include/opendht/def.h
```diff
@@ -42,3 +42,5 @@
 
 // bytes
 #define HASH_LEN 20u
+
+#define OPENDHT_UNUSED_ARG(arg)  (void)arg
```
aberaud: remove

src/connstat.cpp
```diff
@@ -0,0 +1,175 @@
...
+
+    thrd_ = std::thread([this] () { nl_event_loop_thrd(nlsk.get()); });
+}
+
+ConnectivityStatus::~ConnectivityStatus()
+{
+    stop = true;
+
+    nl_socket_modify_cb(nlsk.get(), NL_CB_VALID, NL_CB_CUSTOM, [](nl_msg* msg, void* data) -> int {
+            OPENDHT_UNUSED_ARG(msg);
```
aberaud: simply use `[](nl_msg* /*msg*/, void* data)` in argument list

src/connstat.cpp
```diff
@@ -0,0 +1,175 @@
...
+            if (((ConnectivityStatus*)data)->logger_)
+                ((ConnectivityStatus*)data)->logger_->w("stopping msg processing\n");
```
aberaud: remove log

src/connstat.cpp
```diff
@@ -0,0 +1,175 @@
...
+            return (((ConnectivityStatus*)data)->stop) ? NL_OK : NL_STOP;
+    }, (void*)this);
+
+    nl_socket_modify_cb(nlsk.get(), NL_CB_MSG_IN, NL_CB_CUSTOM, [](nl_msg* msg, void* data) -> int {
+            OPENDHT_UNUSED_ARG(msg);
+            if (((ConnectivityStatus*)data)->logger_)
+                ((ConnectivityStatus*)data)->logger_->w("got incoming msg\n");
```
aberaud: remove

src/connstat.cpp
```diff
@@ -0,0 +1,175 @@
...
+        return (((ConnectivityStatus*)data)->stop) ? NL_OK : NL_STOP;
+    }, (void*)this);
+
+    nl_socket_modify_cb(nlsk.get(), NL_CB_ACK, NL_CB_CUSTOM, [](nl_msg* msg, void* data) -> int {
+            OPENDHT_UNUSED_ARG(msg);
+            if (((ConnectivityStatus*)data)->logger_)
+                ((ConnectivityStatus*)data)->logger_->w("taking care of ACK\n");
```
aberaud: remove

src/connstat.cpp
```diff
@@ -0,0 +1,175 @@
...
+    case RTM_NEWNEIGH:
+    case RTM_DELNEIGH:
+    case RTM_NEWNEIGHTBL:
+        get_neigh_state(msg);
+        break;
+    default:
+        status = NL_SKIP;
+        break;
+    }
+    return status;
```
aberaud: simply return NL_STOP if stopping here ?

src/connstat.cpp
```diff
@@ -0,0 +1,175 @@
...
+
+}
+
+void
+ConnectivityStatus::nl_event_loop_thrd(nl_sock *nlsk)
+{
+    int status;
```
aberaud: status not used

src/connstat.cpp
257: ```NL_STOP``` actually reports error; however for ```NL_CB_VALID``` it wouldn't make a difference.

i moved us to ```NL_CB_MSG_IN``` which means more control; e.g. terminating early since we get to stop early and possibly exit before libnl starts allocate things for it's routines; ack reply for instance.
anyway, this means that we should return ```NL_OK``` so that nllib continues with its routine in case we're not terminating; continuing with next msg for multipart ones or sending back ack to kernel and so on.

a side note, the changes i'm about to push to this are not testing on termination because of the problem we saw the other day.

PR #483, opened 2020-02-09
Add rust async functions

manuels:
Hey, here is a first sketch of who async could be implemented.
I'm not quite sure about the Errors that could be raised, i.e. in what cases does get()'s done callback return ok=false? If the peer is not connected to the network, yet? Maybe we should raise a special Error (right now I am misusing std::io::Error).

Let me know what you think about it?


c/opendht.cpp
```diff
@@ -298,6 +298,24 @@ void dht_runner_bootstrap(dht_runner* r, const char* host, const char* service)
...
 }
 
+void dht_runner_bootstrap2(dht_runner* r, struct sockaddr *addrs[], socklen_t addrs_len[], dht_done_cb done_cb, void* cb_user_data) {
+    auto runner = reinterpret_cast<dht::DhtRunner*>(r);
+
+    std::vector<dht::SockAddr> sa;
+
+    size_t i = 0;
+    while(addrs != nullptr && addrs[i] != nullptr) {
+        sa.push_back(dht::SockAddr(addrs[i], addrs_len[i]));
```
AmarOk1412: emplace_back

rust/examples/dhtnode_async.rs
```diff
@@ -0,0 +1,69 @@
...
+    dht.run_config(1412, config);
+    use std::net::ToSocketAddrs;
+    let addrs = "bootstrap.jami.net:4222".to_socket_addrs().unwrap();
+
+    futures::executor::block_on(async {
+        let r = dht.bootstrap_async(addrs).await;
+
+        println!("Current node id: {}", dht.node_id());
+
+        let mut stream = dht.get_async(&InfoHash::get("bob"));
```
AmarOk1412: Imho, we can do like reqwest crate. Async by default and a sync Dht builder if someone needs to do some sync operations. So we can have DhtRunner::new(); and DhtRunner::blocking::new(), then .get() for both versions (https://docs.rs/reqwest/0.10.1/reqwest/blocking/index.html vs https://docs.rs/reqwest/0.10.1/reqwest/)

If you want I can take some time to modify this :). I think adding _async everywhere is not the best way

AmarOk1412:
For the get, "ok" in the callback will be false if the Dht is not running, this can occurs during a connectivity change too.

AmarOk1412:
Maybe a std::io::NotConnected makes more sense.

manuels:
Hey @AmarOk1412,
I'm a bit busy these days, but I hope I can fix it this weekend.

AmarOk1412:
Np I'm not available till Feb 26 so it will take time :)

PR #492, opened 2020-04-23
net: add updateValues to replace Listen operations

AmarOk1412:
Actually, packets are not retransmitted for listen operations. This
leads to some problems when they are already several values present
at one hash. Ideally, packets should be transmitted inside a request
to avoid packet loss and get all datas.

This patch breaks the compatibility with previous versions

### DIFF
--- a/src/dht.cpp
+++ b/src/dht.cpp
@@ -1200,13 +1200,8 @@ Dht::storageChanged(const InfoHash& id, Storage& st, ValueStorage& v, bool newVa
                 std::vector<Sp<Value>> vals {};
                 vals.push_back(v.data);
                 Blob ntoken = makeToken(node_listeners.first->getAddr(), false);
-                if (l.second.version == 1) {
-                    network_engine.updateValues(node_listeners.first, l.first, id, 0, ntoken, {}, {},
-                            std::move(vals), l.second.query);
-                } else {
-                    network_engine.tellListener(node_listeners.first, l.first, id, 0, ntoken, {}, {},
-                            std::move(vals), l.second.query);
-                }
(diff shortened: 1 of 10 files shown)


src/dht.cpp
```diff
@@ -1287,11 +1292,17 @@ Dht::storageAddListener(const InfoHash& id, const Sp<Node>& node, size_t socket_
...
-                    std::move(vals), query);
+            if (version == 1) {
+                network_engine.updateValues(node, socket_id, id, WANT4 | WANT6, makeToken(node->getAddr(), false),
+                        buckets4.findClosestNodes(id, now, TARGET_NODES), buckets6.findClosestNodes(id, now, TARGET_NODES),
+                        std::move(vals), query);
+            } else {
+                network_engine.tellListener(node, socket_id, id, WANT4 | WANT6, makeToken(node->getAddr(), false),
+                        buckets4.findClosestNodes(id, now, TARGET_NODES), buckets6.findClosestNodes(id, now, TARGET_NODES),
+                        std::move(vals), query);
+            }
```
aberaud, line 1303: move this logic inside tellListener (it should be invisible to Dht except the version number)

tests/dhtrunnertester.cpp
```diff
@@ -167,4 +167,54 @@ DhtRunnerTester::testListen() {
...
+    std::mutex mutex;
+    std::condition_variable cv;
+    std::atomic_uint valueCount(0);
+    unsigned putCount(0);
+    unsigned putOkCount(0);
+
+    std::string data {};
+    for (int i = 0; i < 10000; ++i) {
+        data += "a";
+    }
```
aberaud, line 181: std::string data(10000, 'a');

PR #495, opened 2020-05-04
listener: fully migrate to requests

AmarOk1412:
This patch migrates tellListenerRefreshed and tellListenerExpired
to the new behavior


src/net.h
```diff
@@ -32,7 +32,9 @@ enum class MessageType {
     Listen,
     ValueData,
     ValueUpdate,
-    UpdateValue
+    UpdateValue,
+    RefreshValue,
+    ExpireValue
```
aberaud, line 37: could we use a single message type for the 3 ? (since it goes to the socket anyway)

AmarOk1412: Indeed UpdateValues is definitely enough to do all the job and avoid to create one more version

PR #498, opened 2020-05-11
Proxy: persist ops

aberaud:


AmarOk1412:
hmm from my tests (used the code on proxy3, port 97 and use it in my android client). This doesn't seems to work

```
[514591.076358] Using push server for notifications: localhost:8088
[514591.078410] Error loading state from file: std::bad_cast
```

AmarOk1412:
Now it works :)

PR #527, opened 2020-11-04
http: use local cache to store pre-decoded certs

atraczyk:
- pre-decode certs from DER->PEM and store them so http connections
  can add them the ssl context x509 store
- fixes the windows build by adding the crypto/x509 header


aberaud:
merged seperatly

PR #551, opened 2021-04-27
http: add support for TCP keep alive and handle broken pipes

AmarOk1412:


src/http.cpp
```diff
@@ -561,12 +561,64 @@ void
...
+        return;
+    }
+    auto& base = ssl_socket_? ssl_socket_->lowest_layer() : *socket_;
+
+    ConnectHandlerCb wcb = [&base, cb=std::move(cb)](const asio::error_code& ec, const asio::ip::tcp::endpoint& endpoint) {
+        if (!ec) {
+            auto socket = base.native_handle();
+            // Once connected, set a keep alive on the TCP socket with 30 seconds delay
+            // This will generate broken pipes as soon as possible.
+            // Note this as to be done once connected to have a valid native_handle()
```
aberaud, line 575: has

src/http.cpp
```diff
@@ -561,12 +561,64 @@ void
...
+            std::string cnt_str = std::to_string(cnt);
+            setsockopt(socket, IPPROTO_TCP, TCP_KEEPCNT,
+                    cnt_str.c_str(), sizeof(cnt_str));
+#endif
+#else
+            uint32_t val = 1;
+            setsockopt(socket, SOL_SOCKET, SO_KEEPALIVE, &val, sizeof(uint32_t));
+            uint32_t start = 30;
+            uint32_t interval = 30;
+            uint32_t cnt = 1;
```
aberaud, line 602: defined after their use above

AmarOk1412: Bad cleanup :/

PR #554, opened 2021-05-25
Long public key

aberaud:


AmarOk1412:
Looks good to me. didn't test with the daemon yet

AmarOk1412 (approved):

PR #556, opened 2021-05-27
Cleanups

aberaud:


include/opendht/sockaddr.h
```diff
@@ -68,7 +68,7 @@ class OPENDHT_PUBLIC SockAddr {
      * Build from existing address.
      */
     SockAddr(const sockaddr* sa, socklen_t length) {
-        if (length > sizeof(sockaddr_storage))
+        if (length > (socklen_t)sizeof(sockaddr_storage))
```
AmarOk1412, line 71: oh a C-style cast

src/dhtrunner.cpp
```diff
@@ -119,22 +119,24 @@ DhtRunner::run(const Config& config, Context&& context)
...
-                if (context.logger)
-                    context.logger->d("[runner %p] Using IPv4 port %hu from saved configuration", this, port);
-                if (local4.getPort() == 0)
+                if (local4.getPort() == 0) {
+                    if (context.logger)
+                        context.logger->d("[runner %p] Using IPv4 port %hu from saved configuration", this, port);
                     local4.setPort(port);
+                }
             }
             if (inConfig >> port) {
```
AmarOk1412, line 134: In fact we can do this instruction only if local6.getPort() == 0

include/opendht/sockaddr.h
aberaud: C-style casts are the best ;)

src/dhtrunner.cpp
aberaud: Indeed, but keeping like that since it's an uncommon case, and it's cleaner to have the same logic for both.

PR #560, opened 2021-06-10
thread_pool: Add `onThreadCreated` method

oldiob:
Allows callback on thread creation.

Example:
```c++
dht::ThreadPool::io().onThreadCreated([] {
        sip_utils::register_thead();
});
```

### DIFF
--- a/include/opendht/thread_pool.h
+++ b/include/opendht/thread_pool.h
@@ -59,6 +59,10 @@ public:
         return get(std::move(cb));
     }
 
+    void onThreadCreated(std::function<void(void)>&& cb) {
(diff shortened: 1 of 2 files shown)


aberaud:
I see the point but I don't think this is the right approach.
The thread pool is global, so many different components/users/libraries might want to register various competing initializers, so it would need a list of initializers. But what if a component registers an initializer multiple times? It would need to be able to unregister initializers, and it becomes a heavy and inefficient mechanism for something (thread_local variables) that is already handled natively just fine. Instead of just constructing thread_local object on first use, every thread in the pool would have initialized every single thread_local variable of every user...

Issue #575, opened 2022-02-14
dhtnode -b with unresponsive node goes crazy after a while with pings


ghenry:
Hi all,

Using:

db000a2f8849584f84a26cba8583a188741e0418 (HEAD -> master, tag: 2.3.2, origin/master, origin/HEAD)

I am doing:

`./dhtnode -b bootstrap.sentrypeer.org:4222 -v`

and l see this initially:

```
[1221715.404445] [runner 0x1c8a550] state changed to Running
[1221715.404571] DHT node initialised with ID a6e97a004338654a8814e9bf0d1daa22de06316f
Bootstrap: bootstrap.sentrypeer.org:4222
OpenDHT node a6e97a004338654a8814e9bf0d1daa22de06316f running on port 54559
 (type 'h' or 'help' for a list of possible commands)

[1221715.404737] Bootstraping
>> [1221715.406222] Sending ping to 144.21.55.36:4222
[1221715.406301] Connectivity changed: IPv4: connecting, IPv6: disconnected
[1221723.761423] Bootstraping
[1221723.762319] Sending ping to 144.21.55.36:4222
[1221723.762394] Connectivity changed: IPv4: disconnected, IPv6: disconnected
[1221725.406398] Connectivity changed: IPv4: connecting, IPv6: disconnected
[1221733.481673] Bootstraping
[1221733.482270] Sending ping to 144.21.55.36:4222
[1221733.482346] Connectivity changed: IPv4: disconnected, IPv6: disconnected
```
I'm playing with the Oracle Cloud firewall and don't have port 4222 UDP open yet. After a while (which I'll confirm the time duration), this scrolls past very fast:

```
[1221589.020516] Bootstraping
[1221589.020864] Sending ping to 144.21.55.36:4222
[1221589.020913] Bootstraping
[1221589.021284] Sending ping to 144.21.55.36:4222
[1221589.021346] Bootstraping
[1221589.021636] Sending ping to 144.21.55.36:4222
[1221589.021676] Bootstraping
[1221589.021959] Sending ping to 144.21.55.36:4222
[1221589.022019] Bootstraping
[1221589.022331] Sending ping to 144.21.55.36:4222
[1221589.022378] Bootstraping
[1221589.022661] Sending ping to 144.21.55.36:4222
[1221589.022729] Bootstraping
[1221589.023005] Sending ping to 144.21.55.36:4222
[1221589.023060] Bootstraping
[...]

PR #576, opened 2022-02-16
Bootstrap: fix exponential backoff

aberaud:


src/dht.cpp
```diff
@@ -2052,10 +2056,27 @@ Dht::onDisconnected()
...
+}
+
+void
+Dht::stopBootstrap()
+{
+    if (bootstrapJob) {
+        bootstrapJob->cancel();
+        bootstrapJob.reset();
+    }
+    bootstrap_period = std::chrono::seconds(10);
```
AmarOk1412, line 2077: why not using BOOTSTRAP_PERIOD?

src/dht.cpp
```diff
@@ -2052,10 +2056,27 @@ Dht::onDisconnected()
     }
     if (bootstrapJob)
         bootstrapJob->cancel();
-    bootstrapJob = scheduler.add(scheduler.time() + bootstrap_period, std::bind(&Dht::onDisconnected, this));
+    bootstrapJob = scheduler.add(scheduler.time() + bootstrap_period, std::bind(&Dht::bootstrap, this));
```
AmarOk1412, line 2059: I still not really understand how it can cause https://github.com/savoirfairelinux/opendht/issues/575

I mean, the bootstrap_period should never be < 10 secs.

The only thing I see, ::time() may return an old value causing now + 10 to be < to the actual now() (and causing an infinite loop), maybe a missing syncTime() ? Maybe add syncTime in ::bootstrap() ?

src/dht.cpp

src/dht.cpp
aberaud: Main issue is fixed by this commit. Overflow after 252 years to be handled in another commit.

aberaud: done

AmarOk1412 (approved):

PR #580, opened 2022-02-19
dht: avoid job & search leak

aberaud:


src/dht.cpp
```diff
@@ -2563,8 +2567,15 @@ Dht::storageRefresh(const InfoHash& id, Value::Id vid)
         }
 
         auto expiration = s->second.refresh(now, vid, types);
```
AmarOk1412, line 2569: little comment but for clarity we can use "auto [storage, expiration]" and avoid a lot of .first after this

aberaud: OpenDHT is still C++14

src/dht.cpp
```diff
@@ -825,7 +827,7 @@ Dht::search(const InfoHash& id, sa_family_t af, GetCallback gcb, QueryCallback q
         sr->expired = false;
         sr->nodes.clear();
         sr->nodes.reserve(SEARCH_NODES+1);
-        sr->nextSearchStep = scheduler.add(time_point::max(), std::bind(&Dht::searchStep, this, sr));
+        sr->nextSearchStep = scheduler.add(time_point::max(), std::bind(&Dht::searchStep, this, std::weak_ptr<Search>(sr)));
```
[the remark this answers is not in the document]
aberaud: There is no ws in this function

AmarOk1412 (approved):

PR #594, opened 2022-03-25
Fixcppcheck

lucassdiass:
Cppcheck indicates some errors:
opendht/src/dht.cpp:837:5: error: Returning/dereferencing 'sr' after it is deallocated / released [deallocret]
    return sr;
    ^
opendht/src/dht.cpp:827:119: note: Returning/dereferencing 'sr' after it is deallocated / released
        sr->nextSearchStep = scheduler.add(time_point::max(), std::bind(&Dht::searchStep, this, std::weak_ptr<Search>(sr)));
                                                                                                                      ^
opendht/src/dht.cpp:837:5: note: Returning/dereferencing 'sr' after it is deallocated / released
    return sr;
    ^
opendht/src/search.h:295:29: error: Iterators of different containers are used together. [mismatchingContainers]
            }) != status.end();

This change fixes the mismatchingContainers


lucassdiass:
This change fixes the mismatchingContainers and deallocret

aberaud:
Thanks!
A different mismatchingContainers fix was merged separately.
"deallocret" seems to be a false positive?

lucassdiass:
I think it is a false positive because searchStep verifies if weak_ptr.lock() is expired or not. I've seen the fix of mismatchingContainers and it is better than my change. So, I will close this pull request.

PR #598, opened 2022-03-28
Support `risc-v` architecture

SpriteOvO:
Currently this project cannot be compiled for `risc-v` architecture, `gcc` compiler raises an error:

```
/usr/bin/ld: libopendht.so.2.3.5: undefined reference to `__atomic_exchange_1'
```

This is because the `risc-v` architecture does not support 1-byte and 2-byte lock-free atomic operations.

By passing the `-latomic` option, `gcc` will statically link the `libatomic` library. Inside that library, the `std::atomic<bool>` used in the project will fallback to the mutex implementation, which is allowed by C++ standard.

Ref: https://github.com/riscv-collab/riscv-gnu-toolchain/issues/183#issuecomment-253721765

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -320,6 +320,9 @@ if (OPENDHT_STATIC)
             PUBLIC ${CMAKE_THREAD_LIBS_INIT} ${GNUTLS_LIBRARIES} ${Nettle_STATIC_LIBRARIES}
                    ${Jsoncpp_STATIC_LIBRARIES} ${FMT_LIBRARY} ${HTTP_PARSER_LIBRARY}
                    ${OPENSSL_STATIC_LIBRARIES})
+        if (NOT HAVE_CXX_ATOMICS_RISC_V_WITHOUT_LIB)
+            target_link_libraries(opendht-static PUBLIC "atomic")
+        endif ()
     else ()
         if (OPENDHT_TOOLS)
             function (add_obj_lib name libfile)
@@ -393,6 +396,9 @@ if (OPENDHT_SHARED)
             PRIVATE ${GNUTLS_LIBRARIES} ${Nettle_LIBRARIES}
                     ${Jsoncpp_LIBRARIES}
                     ${FMT_LIBRARY} ${HTTP_PARSER_LIBRARY} ${argon2_LIBRARIES})
+        if (NOT HAVE_CXX_ATOMICS_RISC_V_WITHOUT_LIB)
+            target_link_libraries(opendht PUBLIC "atomic")
+        endif ()
     endif ()
 
     install (TARGETS opendht DESTINATION ${CMAKE_INSTALL_LIBDIR} EXPORT opendht)
@@ -497,6 +503,9 @@ if (OPENDHT_TESTS)
        ${GNUTLS_LIBRARIES}
        ${Jsoncpp_LIBRARIES}
     )
+    if (NOT HAVE_CXX_ATOMICS_RISC_V_WITHOUT_LIB)
+        target_link_libraries(opendht_unit_tests "atomic")
+    endif ()
     if (OPENDHT_PROXY_OPENSSL)
         target_link_libraries(opendht_unit_tests ${OPENSSL_LIBRARIES})
     endif()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -320,8 +320,8 @@ if (OPENDHT_STATIC)
             PUBLIC ${CMAKE_THREAD_LIBS_INIT} ${GNUTLS_LIBRARIES} ${Nettle_STATIC_LIBRARIES}
                    ${Jsoncpp_STATIC_LIBRARIES} ${FMT_LIBRARY} ${HTTP_PARSER_LIBRARY}
                    ${OPENSSL_STATIC_LIBRARIES})
-        if (NOT HAVE_CXX_ATOMICS_RISC_V_WITHOUT_LIB)
-            target_link_libraries(opendht-static PUBLIC "atomic")
+        if (NOT HAVE_CXX_ATOMICS_WITHOUT_LIB)
+            target_link_libraries(opendht-static PUBLIC atomic)
         endif ()
     else ()
         if (OPENDHT_TOOLS)
@@ -396,8 +396,8 @@ if (OPENDHT_SHARED)
             PRIVATE ${GNUTLS_LIBRARIES} ${Nettle_LIBRARIES}
                     ${Jsoncpp_LIBRARIES}
                     ${FMT_LIBRARY} ${HTTP_PARSER_LIBRARY} ${argon2_LIBRARIES})
-        if (NOT HAVE_CXX_ATOMICS_RISC_V_WITHOUT_LIB)
-            target_link_libraries(opendht PUBLIC "atomic")
+        if (NOT HAVE_CXX_ATOMICS_WITHOUT_LIB)
+            target_link_libraries(opendht PUBLIC atomic)
         endif ()
     endif ()
 
@@ -503,8 +503,8 @@ if (OPENDHT_TESTS)
        ${GNUTLS_LIBRARIES}
        ${Jsoncpp_LIBRARIES}
     )
-    if (NOT HAVE_CXX_ATOMICS_RISC_V_WITHOUT_LIB)
-        target_link_libraries(opendht_unit_tests "atomic")
+    if (NOT HAVE_CXX_ATOMICS_WITHOUT_LIB)
+        target_link_libraries(opendht_unit_tests atomic)
     endif ()
     if (OPENDHT_PROXY_OPENSSL)
         target_link_libraries(opendht_unit_tests ${OPENSSL_LIBRARIES})
--- a/cmake/CheckAtomic.cmake
+++ b/cmake/CheckAtomic.cmake
@@ -34,6 +34,21 @@ int main() {
   set(CMAKE_REQUIRED_FLAGS ${OLD_CMAKE_REQUIRED_FLAGS})
 endfunction(check_working_cxx_atomics64)
 
+function(check_working_cxx_atomics_for_risc_v varname)
+  set(OLD_CMAKE_REQUIRED_FLAGS ${CMAKE_REQUIRED_FLAGS})
+  set(CMAKE_REQUIRED_FLAGS "${CMAKE_REQUIRED_FLAGS} -std=c++11")
+  CHECK_CXX_SOURCE_COMPILES("
+#include <atomic>
+std::atomic<uint8_t> x;
+std::atomic<uint16_t> y;
+int main() {
+  x++;
+  y++;
+  return 0;
+}
+" ${varname})
+  set(CMAKE_REQUIRED_FLAGS ${OLD_CMAKE_REQUIRED_FLAGS})
+endfunction(check_working_cxx_atomics_for_risc_v)
 
 # This isn't necessary on MSVC, so avoid command-line switch annoyance
 # by only running on GCC-like hosts.
@@ -73,6 +88,22 @@ if(NOT HAVE_CXX_ATOMICS64_WITHOUT_LIB)
     endif()
   else()
     message(FATAL_ERROR "Host compiler appears to require libatomic for 64-bit operations, but cannot find it.")
+  endif()
+endif()
+
+# Check for RISC-V atomic operations.
+if(MSVC)
+  set(HAVE_CXX_ATOMICS_RISC_V_WITHOUT_LIB True)
+else()
+  check_working_cxx_atomics_for_risc_v(HAVE_CXX_ATOMICS_RISC_V_WITHOUT_LIB)
+endif()
+
+# If not, check if the library exists, and atomics work with it.
+if(NOT HAVE_CXX_ATOMICS_RISC_V_WITHOUT_LIB)
+  list(APPEND CMAKE_REQUIRED_LIBRARIES "atomic")
+  check_working_cxx_atomics_for_risc_v(HAVE_CXX_ATOMICS_RISC_V_WITH_LIB)
+  if (NOT HAVE_CXX_ATOMICS_RISC_V_WITH_LIB)
+    message(FATAL_ERROR "Host compiler must support RISC-V std::atomic!")
   endif()
 endif()
 
--- a/cmake/CheckAtomic.cmake
+++ b/cmake/CheckAtomic.cmake
@@ -12,8 +12,12 @@ function(check_working_cxx_atomics varname)
   CHECK_CXX_SOURCE_COMPILES("
 #include <atomic>
 std::atomic<int> x;
+std::atomic<short> y;
+std::atomic<char> z;
 int main() {
-  return x;
+  ++z;
+  ++y;
+  return ++x;
 }
 " ${varname})
   set(CMAKE_REQUIRED_FLAGS ${OLD_CMAKE_REQUIRED_FLAGS})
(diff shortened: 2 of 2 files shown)


r-value:
It would be better to check if libatomic is necessary before linking against it.

The `CheckAtomic` module from llvm could be a reference: https://github.com/llvm/llvm-project/blob/main/llvm/cmake/modules/CheckAtomic.cmake

SpriteOvO:
I just noticed that [`CheckAtomic.cmake`](https://github.com/savoirfairelinux/opendht/blob/master/cmake/CheckAtomic.cmake) was included in the project over a year ago, but it doesn't seem to be working and I'm doing some investigation.

SpriteOvO:
Updated. Link `libatomic` only if the atomic tests fail, to avoid other architectures linking it unnecessarily.

CMakeLists.txt
```diff
@@ -382,14 +382,14 @@ if (OPENDHT_SHARED)
     target_include_directories(opendht SYSTEM PRIVATE ${argon2_INCLUDE_DIRS})
     if (APPLE)
         target_link_libraries(opendht
-            PUBLIC ${CMAKE_THREAD_LIBS_INIT} ${OPENSSL_LIBRARIES}
+            PUBLIC ${CMAKE_REQUIRED_LIBRARIES} ${CMAKE_THREAD_LIBS_INIT} ${OPENSSL_LIBRARIES}
```
r-value, line 385: This branch can't be affected by RISC-V related chagnes, too

CMakeLists.txt
```diff
@@ -364,7 +364,7 @@ if (OPENDHT_STATIC)
             ${opendht_HEADERS}
             ${obj_libs}
         )
-        target_link_libraries(opendht-static PUBLIC ${Win32_STATIC_LIBRARIES} ${Win32_IMPORT_LIBRARIES})
+        target_link_libraries(opendht-static PUBLIC ${CMAKE_REQUIRED_LIBRARIES} ${Win32_STATIC_LIBRARIES} ${Win32_IMPORT_LIBRARIES})
```
r-value, line 367: This branch is `MSVC` which shouldn't be affected by RISC-V related changes.

CMakeLists.txt
```diff
@@ -382,14 +382,14 @@ if (OPENDHT_SHARED)
...
-            PUBLIC ${CMAKE_THREAD_LIBS_INIT} ${OPENSSL_LIBRARIES}
+            PUBLIC ${CMAKE_REQUIRED_LIBRARIES} ${CMAKE_THREAD_LIBS_INIT} ${OPENSSL_LIBRARIES}
             PRIVATE ${GNUTLS_LIBRARIES} ${Nettle_LIBRARIES}
                     ${Jsoncpp_LIBRARIES}
                     ${FMT_LIBRARY} ${HTTP_PARSER_LIBRARY} ${argon2_LIBRARIES}
             SYSTEM "-framework CoreFoundation" "-framework Security")
     else ()
         target_link_libraries(opendht
-            PUBLIC ${CMAKE_THREAD_LIBS_INIT} ${OPENSSL_LIBRARIES}
+            PUBLIC ${CMAKE_REQUIRED_LIBRARIES} ${CMAKE_THREAD_LIBS_INIT} ${OPENSSL_LIBRARIES}
```
r-value, line 392: Using `CMAKE_REQUIRED_LIBRARIES` may not be a good idea here.

r-value (changes requested):

cmake/CheckAtomic.cmake
```diff
@@ -34,23 +38,17 @@ int main() {
...
 
-
 # This isn't necessary on MSVC, so avoid command-line switch annoyance
 # by only running on GCC-like hosts.
 if (LLVM_COMPILER_IS_GCC_COMPATIBLE)
   # First check if atomics work without the library.
   check_working_cxx_atomics(HAVE_CXX_ATOMICS_WITHOUT_LIB)
   # If not, check if the library exists, and atomics work with it.
   if(NOT HAVE_CXX_ATOMICS_WITHOUT_LIB)
-    check_library_exists(atomic __atomic_fetch_add_4 "" HAVE_LIBATOMIC)
```
r-value, line 45: I'd like to make some further explanation here.

`CHECK_LIBRARY_EXISTS` won't work if the function to look for is built-in. The test program will cause a signature conflict thus fail to compile, leading to a false negative.

r-value, line 45: I'd like to make some further explanation here.

`CHECK_LIBRARY_EXISTS` won't work if the function to look for is built-in. The test program will cause a signature conflict thus fail to compile, leading to a false negative.

aberaud:
I reverted the changes in CMakeLists because adding the library is supposed to be automatic through CMAKE_REQUIRED_LIBRARIES if I understand correctly.
After the merge of this commit, the build was failing on macOS because libatomic couldn't be found.

SpriteOvO:
@aberaud In fact, in my previous tests, `CMake` does not automatically link to libraries in `CMAKE_REQUIRED_LIBRARIES`, and I have not found documentation describing this behavior. If we search in the `CMake` repository, we can find some `set(CMAKE_REQUIRED_LIBRARIES xxx)` usage, so we presume that the variable is volatile.

Could we get some error messages on macOS? I didn't find it from the CI. This change is not expected to affect macOS, maybe there are some bugs and I will try to fix it :)

SpriteOvO:
> Could we get some error messages on macOS?

A `CMakeError.log` file would be better.

aberaud:
Looking at various examples as well as the original CheckAtomic from llvm, it seems like using CMAKE_REQUIRED_LIBRARIES is how people do it:
https://stackoverflow.com/questions/69281559/cmake-library-order-in-cmake-required-libraries-to-test-a-minimal-program-while

The error on macOS after the PR was merged:
Note that the same error happens when linking the first executable no matter what build options are used.

cmake:
```
% cmake -DOPENDHT_C=On -DOPENDHT_TESTS=On -DCMAKE_BUILD_TYPE=Release -DCMAKE_INTERPROCEDURAL_OPTIMIZATION=On ..
-- The C compiler identification is AppleClang 13.1.6.13160021
-- The CXX compiler identification is AppleClang 13.1.6.13160021
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /Library/Developer/CommandLineTools/usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /Library/Developer/CommandLineTools/usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Found PkgConfig: /opt/homebrew/bin/pkg-config (found version "0.29.2") 
-- Performing Test HAVE_CXX_ATOMICS64_WITHOUT_LIB
-- Performing Test HAVE_CXX_ATOMICS64_WITHOUT_LIB - Success
-- Performing Test LLVM_HAS_ATOMICS
-- Performing Test LLVM_HAS_ATOMICS - Success
-- Could NOT find Doxygen (missing: DOXYGEN_EXECUTABLE) 
-- Looking for pthread.h
-- Looking for pthread.h - found
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Success
-- Found Threads: TRUE  
-- Found GnuTLS: /opt/homebrew/Cellar/gnutls/3.7.4/lib/libgnutls.dylib (found suitable version "3.7.4", minimum required is "3.3") 
-- Checking for one of the modules 'nettle'
-- Found Readline: /Library/Developer/CommandLineTools/SDKs/MacOSX12.3.sdk/usr/lib/libreadline.tbd (Required is at least version "6") 
-- Checking for one of the modules 'libargon2'
-- Checking for one of the modules 'jsoncpp'
-- Checking for one of the modules 'cppunit'
-- Configuring done
-- Generating done
-- Build files have been written to: opendht/build
```

build:
```
% make VERBOSE=1
...
[ 82%] Linking CXX executable opendht_unit_tests
/opt/homebrew/Cellar/cmake/3.23.0/bin/cmake -E cmake_link_script CMakeFiles/opendht_unit_tests.dir/link.txt --verbose=1
/Library/Developer/CommandLineTools/usr/bin/c++  -Wno-return-type -Wall -Wextra -Wnon-virtual-dtor -pedantic-errors -fvisibility=hidden -DMSGPACK_NO_BOOST -DMSGPACK_DISABLE_LEGACY_NIL -DMSGPACK_DISABLE_LEGACY_CONVERT -O3 -DNDEBUG -flto=thin -arch arm64 -isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX12.3.sdk -mmacosx-version-min=12.2 -Wl,-search_paths_first -Wl,-headerpad_max_install_names CMakeFiles/opendht_unit_tests.dir/tests/tests_runner.cpp.o CMakeFiles/opendht_unit_tests.dir/tests/infohashtester.cpp.o CMakeFiles/opendht_unit_tests.dir/tests/valuetester.cpp.o CMakeFiles/opendht_unit_tests.dir/tests/cryptotester.cpp.o CMakeFiles/opendht_unit_tests.dir/tests/dhtrunnertester.cpp.o CMakeFiles/opendht_unit_tests.dir/tests/threadpooltester.cpp.o CMakeFiles/opendht_unit_tests.dir/tests/peerdiscoverytester.cpp.o -o opendht_unit_tests   -L/opt/homebrew/Cellar/nettle/3.7.3/lib  -L/opt/homebrew/Cellar/argon2/20190702_1/lib  -L/opt/homebrew/Cellar/cppunit/1.15.1/lib  -Wl,-rpath,/opt/homebrew/Cellar/nettle/3.7.3/lib -Wl,-rpath,/opt/homebrew/Cellar/argon2/20190702_1/lib -Wl,-rpath,/opt/homebrew/Cellar/cppunit/1.15.1/lib -Wl,-rpath,opendht/build libopendht.2.4.0.dylib -lcppunit /opt/homebrew/Cellar/gnutls/3.7.4/lib/libgnutls.dylib -latomic 
ld: library not found for -latomic
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make[2]: *** [opendht_unit_tests] Error 1
make[1]: *** [CMakeFiles/opendht_unit_tests.dir/all] Error 2
make: *** [all] Error 2
```

SpriteOvO:
@aberaud I did some investigation and I now understand how everything is going on.

---

> Looking at various examples as well as the original CheckAtomic from llvm, it seems like using CMAKE_REQUIRED_LIBRARIES is how people do it: [stackoverflow.com/questions/69281559/cmake-library-order-in-cmake-required-libraries-to-test-a-minimal-program-while](https://stackoverflow.com/questions/69281559/cmake-library-order-in-cmake-required-libraries-to-test-a-minimal-program-while)

I wrote a minimal example for testing and the result is that `CMAKE_REQUIRED_LIBRARIES` does not work for a target. It works for `CHECK_CXX_SOURCE_COMPILES`.

Quoted from [the documentation of `check_cxx_source_compiles`](https://cmake.org/cmake/help/latest/module/CheckCXXSourceCompiles.html):

> `CMAKE_REQUIRED_LIBRARIES`
>
> A [;-list](https://cmake.org/cmake/help/latest/manual/cmake-language.7.html#cmake-language-lists) of libraries to add to the link command. These can be the name of system libraries or they can be [Imported Targets](https://cmake.org/cmake/help/latest/manual/cmake-buildsystem.7.html#imported-targets) (see [try_compile()](https://cmake.org/cmake/help/latest/command/try_compile.html#command:try_compile) for further details).

The code in the StackOverflow question you mentioned also implies my point.

Quoted from [the StackOverflow question code](https://stackoverflow.com/questions/69281559/cmake-library-order-in-cmake-required-libraries-to-test-a-minimal-program-while):

```cmake
# Before setting CMAKE_REQUIRED_FLAGS, we preserve the current state
cmake_push_check_state()

set(CMAKE_REQUIRED_LIBRARIES "-L${atomic_lib_dir} -latomic")
check_cxx_source_compiles("${ATOMIC32_TEST_CODE}" atomic32_test_with_atomic_linking)
check_cxx_source_compiles("${ATOMIC64_TEST_CODE}" atomic64_test_with_atomic_linking)

cmake_pop_check_state()
```

I treat the `CMAKE_REQUIRED_FLAGS` in the comment as a typo, which should be `CMAKE_REQUIRED_LIBRARIES`, because `CMAKE_REQUIRED_FLAGS` is not used anywhere in his code.

Quoted from [the documentation of `cmake_push_check_state`](https://cmake.org/cmake/help/latest/module/CMakePushCheckState.html):

> This module defines three macros: `CMAKE_PUSH_CHECK_STATE()` `CMAKE_POP_CHECK_STATE()` and `CMAKE_RESET_CHECK_STATE()` **These macros can be used to save, restore and reset (i.e., clear contents) the state of the variables** `CMAKE_REQUIRED_FLAGS`, `CMAKE_REQUIRED_DEFINITIONS`, `CMAKE_REQUIRED_LINK_OPTIONS`, **`CMAKE_REQUIRED_LIBRARIES`**, `CMAKE_REQUIRED_INCLUDES` and `CMAKE_EXTRA_INCLUDE_FILES` used by the various Check-files coming with CMake, like e.g. `check_function_exists()` etc. The variable contents are pushed on a stack, pushing multiple times is supported. This is useful e.g. when executing such tests in a Find-module, where they have to be set, but after the Find-module has been executed they should have the same value as they had before.
>
> `CMAKE_PUSH_CHECK_STATE()` macro receives optional argument `RESET`. Whether it's specified, `CMAKE_PUSH_CHECK_STATE()` will set all `CMAKE_REQUIRED_*` variables to empty values, same as `CMAKE_RESET_CHECK_STATE()` call will do.

---

> The error on macOS:

This is because the `LLVM_COMPILER_IS_GCC_COMPATIBLE` variable in the `CheckAtomic.cmake` file is not working properly.

Let's take a look at the code snippet:

```cmake
# This isn't necessary on MSVC, so avoid command-line switch annoyance
# by only running on GCC-like hosts.
if (LLVM_COMPILER_IS_GCC_COMPATIBLE)
  # First check if atomics work without the library.
  check_working_cxx_atomics(HAVE_CXX_ATOMICS_WITHOUT_LIB)
  # If not, check if the library exists, and atomics work with it.
  if(NOT HAVE_CXX_ATOMICS_WITHOUT_LIB)
    message("ATOMIC_DEBUG: 1 FAIL")
    list(APPEND CMAKE_REQUIRED_LIBRARIES "atomic")
    check_working_cxx_atomics(HAVE_CXX_ATOMICS_WITH_LIB)
    if (NOT HAVE_CXX_ATOMICS_WITH_LIB)
      message(FATAL_ERROR "Host compiler must support std::atomic!")
    else()
      message("ATOMIC_DEBUG: 1 DONE")
    endif()
  endif()
endif()
```

The `LLVM_COMPILER_IS_GCC_COMPATIBLE` variable is actually set from [`DetermineGCCCompatible.cmake`](https://github.com/llvm/llvm-project/blob/main/llvm/cmake/modules/DetermineGCCCompatible.cmake) file, obviously it does not exist in this repo.

When `LLVM_COMPILER_IS_GCC_COMPATIBLE` is not set (is `OFF`), so the atomic check are not performed, so `HAVE_CXX_ATOMICS_WITHOUT_LIB` is always `OFF`, so

```cmake
if (NOT HAVE_CXX_ATOMICS_WITHOUT_LIB)
    target_link_libraries(opendht PUBLIC atomic)
endif ()
```

always links to `libatomic` for the target.

Here's an same bug report from another project: [CheckAtomic doesn't set LLVM_COMPILER_IS_GCC_COMPATIBLE and fails check on Apple Clang](https://bugs.kde.org/show_bug.cgi?id=444247)

---

I would wait for the macOS CI of this repo to work properly and then open a new PR. If you have any questions let me know :)

aberaud:
Thanks for the detailed analysis and explanations !
* Shall we then use the generic [link_libraries](https://cmake.org/cmake/help/latest/command/link_libraries.html) instead of per-target with target_link_libraries ?
* The macOS CI job should now be working well enough :)

SpriteOvO:
> * Shall we then use the generic [link_libraries](https://cmake.org/cmake/help/latest/command/link_libraries.html) instead of per-target with target_link_libraries ?

Of course, it should work.

> * The macOS CI job should now be working well enough :)

Nice! I will open a PR later.

PR #601, opened 2022-04-03
Impersonation

lucassdiass:
DHT allows the impersonation attack because method getNode updates the node's addr in following case:
node A (IP address 1.1.1.1) has InfoHash 8ffe0d18748cfc04d1bc0a2e8c3a548e06f3f794 and sends a Ping to node B (IP address 2.2.2.2), then node B sends neighborhood maintenance to 1.1.1.1.
Thus, a malicious node (IP Address 3.3.3.3) discovers the InfoHash 8ffe0d18748cfc04d1bc0a2e8c3a548e06f3f794 from Node A and sends DHT packets to Node B.
Thus, when node B receives a message, it will update the node A address. Then, it will send the neighborhood maintenance to 3.3.3.3 instead 1.1.1.1.
With this change, node A will update the node addr when node B sends a confirmation and it is old. When node B is not old, node A updates the node addr with its actual addr.


aberaud:
Thanks a lot for mentioning the issue and for the patch. 

In the long run, we plan to fix this by using libsodium for protocol-level encryption and node authentication.

The only problem with the current PR is that if the node IP address changes, this node won't receive answers for the next 10 minutes. Nodes' IP addresses can change often enough in normal conditions for this to be an issue.

In general, since OpenDHT is an open network, values received should always be considered potentially malicious.
For this reason, OpenDHT provides an optional data-level encryption and signature layer. Applications can also implement their own encryption/signature mechanisms.

lucassdiass:
I understand and I will wait for this update.
I work on a fix also. In the branch https://github.com/lucassdiass/opendht/tree/PKIpp, I enable support for the PKI to the node authentication using https://github.com/lucassdiass/PKIpp (based on OpenSSL). However, it must be improved. For example, CRL does not work 100%.

PR #615, opened 2022-06-07
dhtrunner: remove support for running both dht and proxy client

aberaud:
If configured to run as a proxy client, no port will be bound and the local dht node
won't be instantiated.

Trying to use the proxy if not compiled will now trigger an exception.


AmarOk1412 (approved):
Code is ok, however 2.4.4 is crashing with Jami (at least on linux)... bisecting..

PR #625, opened 2022-07-04
proxyserver: add application bundle id option

katekm:


src/dht_proxy_server.cpp
```diff
@@ -788,6 +789,9 @@ DhtProxyServer::subscribe(restinio::request_handle_t request,
         }
         auto type = getTypeFromString(root["platform"].asString());
         auto topic = root["topic"].asString();
+        if (topic.empty() && !bundleId_.empty()) {
+            topic = type == PushType::iOS ? bundleId_ : type == PushType::iOSLegacy ? bundleId_  + ".voip" : "";
+        }
```
aberaud, line 794: factorize logic

src/dht_proxy_server.cpp
```diff
@@ -1188,6 +1192,9 @@ DhtProxyServer::put(restinio::request_handle_t request,
                         pput.pushToken = pushToken;
                         pput.clientId = clientId;
                         pput.type = getTypeFromString(platform);
+                        if (topic.empty() && !bundleId_.empty()) {
+                            topic = pput.type == PushType::iOS ? bundleId_ :  pput.type == PushType::iOSLegacy ? bundleId_  + ".voip" : "";
+                        }
```
aberaud, line 1197: factorize logic

aberaud (changes requested):

PR #629, opened 2022-08-11
http: modify timeout behaviour

AmarOk1412:
In the previous version, the timeout for the connection could be
set only after async_connect while the request is sending.
This means if async_connect is slow (it can happen as if there is
an ipv6 endpoint, it can try to resolve it during dozens of seconds
before falling back on ipv4).

In this patch, it moves the timeout callback in the request. This
callback will be set just before async_connect, to be sure that we
are using the callback while connecting.

### DIFF
--- a/include/opendht/http.h
+++ b/include/opendht/http.h
@@ -114,13 +114,13 @@ public:
     void async_read(size_t bytes, BytesHandlerCb cb);
     void async_read_some(size_t bytes, BytesHandlerCb cb);
 
-    void timeout(const std::chrono::seconds timeout, HandlerCb cb = {});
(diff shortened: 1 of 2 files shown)


include/opendht/http.h
```diff
@@ -261,6 +261,11 @@ class OPENDHT_PUBLIC Request : public std::enable_shared_from_this<Request>
         return resolver_->get_url();
     };
 
+    void timeout(const std::chrono::seconds timeout, HandlerCb cb = {}) {
```
aberaud, line 264: either const reference or simple pass-by-value

aberaud (approved):

aberaud (approved):

aberaud (approved):

PR #651, opened 2023-01-27
Fix value edition, add unit test

aberaud:


src/op_cache.cpp
```diff
@@ -28,6 +28,10 @@ OpValueCache::onValuesAdded(const std::vector<Sp<Value>>& vals, const system_clo
         auto viop = values.emplace(v->id, v);
         if (viop.second) {
             newValues.emplace_back(v);
+        } else if (!(*viop.first->second.data == *v)) {
```
AmarOk1412, line 31: Except this line which is a bit dirty, the rest look ok for me (I would add a test checking if an edition with the same content doesn't trigger the chain)

AmarOk1412 (approved):

src/op_cache.cpp
aberaud: cleaned up the comparaison

PR #661, opened 2023-04-10
cmake: fix msvc build

aberaud:


CMakeLists.txt
atraczyk, line 1: BUILD_TESTING uses `pkg_search_module` too, so:
something like on line 437 -> `if (BUILD_TESTING AND NOT MSVC)`

atraczyk (approved):

PR #685, opened 2023-12-07
search: avoid removing node being added

aberaud:


AmarOk1412:
" avoid removing node being added" but pop_back is executed if it's the last node, so it will be still removed

aberaud:
> " avoid removing node being added" but pop_back is executed if it's the last node, so it will be still removed

done

PR #693, opened 2024-03-20
Shutdown on connecting

aberaud:


src/dht_proxy_client.cpp
```diff
@@ -190,10 +190,12 @@ void
 DhtProxyClient::stop()
 {
     if (not isDestroying_.exchange(true)) {
+        std::shared_ptr<Resolver> resolver;
         {
             std::lock_guard<std::mutex> l(resolverLock_);
-            resolver_.reset();
+            resolver = std::move(resolver_);
         }
+        resolver_->cancel();
```
atraczyk, line 198: resolver_ --> resolver

src/dht_proxy_client.cpp
```diff
@@ -190,10 +190,12 @@ void
 DhtProxyClient::stop()
 {
     if (not isDestroying_.exchange(true)) {
+        std::shared_ptr<Resolver> resolver;
```
atraczyk, line 193: will need the http header

atraczyk:
shutdown-while-connecting deadlock seems to be fixed with this stack

PR #694, opened 2024-04-02
provide value IDs for push notifications

atraczyk:


src/dht_proxy_server.cpp
```diff
@@ -1050,6 +1009,46 @@ DhtProxyServer::handleCancelPushListen(const asio::error_code &ec, const std::st
...
+        if (!ids.empty()) ids += ",";
+        ids += std::to_string(value->id);
+    }
+    json["ids"] = ids;
+
+    // If message is expired copy the value ID to the `exp` field.
+    // This is presumably used by the Android notification system.
+    if (expired && values.size() < 2) {
+        json["exp"] = json["ids"];
+    }
```
aberaud, line 1041: do we want to put the id in "ids" if values expired ?

atraczyk: yes, but only for backwards compatibility for iOS

aberaud (approved):

PR #704, opened 2024-05-14
network_utils: use poll instead of select

fsimonfc:
select is limited to file descriptor numbers below FD_SETSIZE (1024), and calling FD_SET with a higher value can cause a buffer overflow. This can be problematic for applications using OpenDHT as a library.

### DIFF
--- a/src/network_utils.cpp
+++ b/src/network_utils.cpp
@@ -25,10 +25,15 @@
 #include <cstring>
 #define close(x) closesocket(x)
 #define write(s, b, f) send(s, b, (int)strlen(b), 0)
+#define _poll(fds, nfds, timeout) WSAPoll(fds, nfds, timeout)
 #else
-#include <sys/select.h>
 #include <fcntl.h>
+#include <poll.h>
+#define _poll(fds, nfds, timeout) poll(fds, nfds, timeout)
 #endif
+
+// number of file descriptors to poll in openSockets()
+#define NUM_FDS 3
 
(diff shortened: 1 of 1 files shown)


AmarOk1412:
Yeay!

src/network_utils.cpp
```diff
@@ -235,23 +237,23 @@ UdpSocket::openSockets(const SockAddr& bind4, const SockAddr& bind6)
 
     running = true;
     rcv_thread = std::thread([this, stop_readfd, ls4=s4, ls6=s6]() mutable {
-        int selectFd = std::max({ls4, ls6, stop_readfd}) + 1;
+        struct pollfd fds[NUM_FDS];
+        for (int i = 0; i < NUM_FDS; i++)
+            fds[i].events = POLLIN;
+        int stop_readfd_index = 0, ls4_index = 1, ls6_index = 2;
```
aberaud, line 243: constexpr unsigned or constexpr size_t

src/network_utils.cpp
```diff
@@ -26,8 +26,10 @@
 #define close(x) closesocket(x)
 #define write(s, b, f) send(s, b, (int)strlen(b), 0)
 #else
-#include <sys/select.h>
 #include <fcntl.h>
+#include <poll.h>
```
aberaud, line 30: what about Windows ?

aberaud (changes requested):
looks mostly good

PR #708, opened 2024-06-07
peer_discovery: add a workaround for Android hotspot

Lanius-collaris:


aberaud (approved):
Thanks

Lanius-collaris:
Sorry. I forgot to modify reDiscover().

AmnaSnene:
@Lanius-collaris  Could you clarify why you only take into account IPv4 interfaces?

Lanius-collaris:
@AmnaSnene On android, we always have a route to dummy0 (e.g. `multicast ff00::/8 dev dummy0 table local proto kernel metric 256 pref medium`), so it's impossible to encounter "No such device" error when setting `asio::ip::multicast::join_group` option for IPv6.
You can try `adb shell ip route get ff08::101` when WLAN, mobile data and hotspot are OFF.

PR #709, opened 2024-06-09
make crypto error messages more readable

lucassdiass:
Add verifications to gnutls calls and throw expcetion with readable messages.
For example, if call loadIdentity and crt file has name different of private key exception message was
Can't get certificate public key ID: ASN1 parser: Value was not found.

However, the real problem is that crt file was not found. Now, for the same problem the exception message will be: Error while reading file.

### DIFF
--- a/src/crypto.cpp
+++ b/src/crypto.cpp
@@ -1152,12 +1152,19 @@ loadIdentity(const std::string &path,const std::string &privkey_password)
     pkStream.close();
     // Create a certificate
     gnutls_x509_crt_t gnuCert;
-    if (gnutls_x509_crt_init(&gnuCert) != GNUTLS_E_SUCCESS)
+    if (gnutls_x509_crt_init(&gnuCert))
         throw std::runtime_error("Failed to initialize gnutls certificate struct");
     gnutls_datum_t crtContent {nullptr, 0};
+
     // Read the certificate file
-    gnutls_load_file((path + ".crt").c_str(), &crtContent);
-    gnutls_x509_crt_import(gnuCert, &crtContent, GNUTLS_X509_FMT_PEM);
+    int err = gnutls_load_file((path + ".crt").c_str(), &crtContent);
+    if (err)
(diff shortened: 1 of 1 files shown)


src/crypto.cpp
```diff
@@ -1192,10 +1199,10 @@ Certificate::generate(const PrivateKey& key, const std::string& name, const Iden
     Certificate ret {cert};
 
     setValidityPeriod(cert, validity <= 0 ? 10 * 365 * 24 * 60 * 60 : validity);
-    if (int err = gnutls_x509_crt_set_key(cert, key.x509_key)) {
+    if (int err = gnutls_x509_crt_set_key(cert, key.x509_key) != GNUTLS_E_SUCCESS) {
```
aberaud, line 1202: would cause err to be 0 or 1.

GNUTLS_E_SUCCESS is defined to be 0: https://gnutls.org/reference/gnutls-gnutls.html#GNUTLS-E-SUCCESS:CAPS

So the simple if construction can be used without the explicit comparison with GNUTLS_E_SUCCESS

src/crypto.cpp
```diff
@@ -1192,10 +1199,10 @@ Certificate::generate(const PrivateKey& key, const std::string& name, const Iden
...
         throw CryptoException(std::string("Error when setting certificate key ") + gnutls_strerror(err));
     }
-    if (int err = gnutls_x509_crt_set_version(cert, 3)) {
+    if (int err = gnutls_x509_crt_set_version(cert, 3) != GNUTLS_E_SUCCESS) {
```
aberaud, line 1205: would cause err to be 0 or 1.

GNUTLS_E_SUCCESS is defined to be 0: https://gnutls.org/reference/gnutls-gnutls.html#GNUTLS-E-SUCCESS:CAPS

So the simple if construction can be used without the explicit comparison with GNUTLS_E_SUCCESS

aberaud (changes requested):
Thanks !

aberaud (approved):

PR #710, opened 2024-06-09
complicate node spoofing

lucassdiass:
DHT allows the impersonation attack because method getNode updates the node's addr in following case: node A (IP address 1.1.1.1) has InfoHash 8ffe0d18748cfc04d1bc0a2e8c3a548e06f3f794 and sends a Ping to node B (IP address 2.2.2.2), then node B sends neighborhood maintenance to 1.1.1.1. Thus, a malicious node (IP Address 3.3.3.3) discovers the InfoHash 8ffe0d18748cfc04d1bc0a2e8c3a548e06f3f794 from Node A and sends DHT packets to Node B.

Thus, when node B receives a message, it will update the node A address. Then, it will send the neighborhood maintenance to 3.3.3.3 instead 1.1.1.1.
With this change, node A will update the node addr when node B sends a confirmation after 1 minute or it is old. When node B is not old, node A updates the node addr with its actual addr.

### DIFF
--- a/include/opendht/node.h
+++ b/include/opendht/node.h
@@ -89,6 +89,9 @@ struct Node {
     bool isRemovable(const time_point& now) const {
         return isExpired() and isOld(now);
     }
+    bool isUpgradeable(const time_point& now) const {
+        return time + NODE_UPDATE_TIME < now;
+    }
 
     NodeExport exportNode() const {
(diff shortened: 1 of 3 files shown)


lucassdiass:
It will avoid flooding also

PR #713, opened 2024-06-21
re-publish and re-discover immediately when the node disconnects

AmnaSnene:


src/dhtrunner.cpp
```diff
@@ -690,8 +696,13 @@ DhtRunner::loop_()
     if (nstatus4 != status4 || nstatus6 != status6) {
         status4 = nstatus4;
         status6 = nstatus6;
-        if (statusCb)
-            statusCb(status4, status6);
+        if (!statusCbs.empty())
```
aberaud, line 699: if not needed

src/dhtrunner.cpp
```diff
@@ -690,8 +696,13 @@ DhtRunner::loop_()
...
+        {
+            for (auto& cb : statusCbs){
+                if (cb)
```
aberaud, line 702: check not needed

src/dhtrunner.cpp
```diff
@@ -180,9 +180,8 @@ DhtRunner::run(const Config& config, Context&& context)
         throw;
     }
 
-    if (context.statusChangedCallback) {
-        statusCb = std::move(context.statusChangedCallback);
-    }
+    statusCbs.clear();
+    statusCbs.emplace_back(std::move(context.statusChangedCallback));
```
aberaud, line 184: only emplace if `statusChangedCallback` is not `{}`.

include/opendht/dhtrunner.h
```diff
@@ -519,7 +519,8 @@ class OPENDHT_PUBLIC DhtRunner {
 
     NodeStatus status4 {NodeStatus::Disconnected},
                status6 {NodeStatus::Disconnected};
-    StatusCallback statusCb {nullptr};
+
+    std::list<StatusCallback> statusCbs {};
```
aberaud, line 523: could be a vector

include/opendht/dhtrunner.h
```diff
@@ -409,7 +409,7 @@ class OPENDHT_PUBLIC DhtRunner {
     void run(const Config& config, Context&& context);
 
     void setOnStatusChanged(StatusCallback&& cb) {
-        statusCb = std::move(cb);
+        statusCbs.emplace_back(std::move(cb));
```
aberaud, line 412: emplace if not empty

src/dhtrunner.cpp
```diff
@@ -264,6 +263,13 @@ DhtRunner::run(const Config& config, Context&& context)
                 }
             }
         }
+        if (config.peer_discovery && config.peer_publish) {
```
aberaud, line 266: what if only `peer_discovery` or `peer_publish` is true ?

aberaud (changes requested):

src/dhtrunner.cpp
AmnaSnene: It should be `config.peer_discovery or config.peer_publish` so i changed it to `if (peerDiscovery_)`.

aberaud (approved):

PR #743, opened 2025-01-25
c/opendht: add missing server_ca, client_identity configs for the proxy

sim590:


sim590:
Was it left aside on purpose?

aberaud (approved):

PR #744, opened 2025-02-02
c: dht_value_with_id_new for setting an ID

sim590:
Necessary for cancelling put values. Otherwise, the user doesn't have the value ID for asking to cancel the permanent put.

### DIFF
--- a/c/opendht.cpp
+++ b/c/opendht.cpp
@@ -132,8 +132,12 @@ void dht_value_set_user_type(dht_value* data, const char* user_type) {
     (*reinterpret_cast<ValueSp*>(data))->user_type = user_type;
 }
 
+dht_value* dht_value_with_id_new(const uint8_t* data, size_t size, uint64_t value_id) {
+    return reinterpret_cast<dht_value*>(new ValueSp(std::make_shared<dht::Value>(dht::ValueType::USER_DATA.id, data, size, value_id)));
+}
+
 dht_value* dht_value_new(const uint8_t* data, size_t size) {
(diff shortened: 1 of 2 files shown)


sim590:
Unless you have another suggestion? Am I missing something?

aberaud (approved):

aberaud:
Thanks for the patch, it adds a feature that was not exposed before with the C API so I merged it.

Note however that this is not required to obtain the value id:
The value id can also be read from the value object as soon as the put callback completes.

To show how this can be done with the C API, I made a patch for dhtcnode to display the value id on put completion:
https://github.com/savoirfairelinux/opendht/pull/745/files

Best regards

sim590:
That's true. I did overlook that the _userdata_ pointer and the donecallback allowed doing this.

However, I think it would not have worked well in the [Haskell layer](https://github.com/sim590/opendht-hs/) as this would have forced me to expose a `DoneCallback` type that would have had a signature including the Value that was put and that choice would have looked arbitrary and out-of-place considering that only the `put` callback would have had this and not the others. In fact, in my design, I kept the callbacks signatures simple by omitting the `userdata` parameter from the exposed Haskell callbacks since I thought that this was done only in order to have an equivalent callback design to the C++ callbacks with variable capture capability. I see now that it also let the user act on the value pointer which is neat for the C users. However, in Haskell, I'm not exposing any type that hides pointers. I try to free all pointers as soon as possible, so it's unfortunately not possible for me to use that. Therefore, I think that my contribution to the C layer is actually needed for my design.

I thought also about returning the ID through the `put` operation, but since that would have been a breakage of API, I thought it was not desirable.

Thanks for merging even though what I did is not necessary for your design of the C layer!

PR #750, opened 2025-04-22
Added and tested build scenarios for Windows (MSYS2, MSVC static)

twdragon:
- Based on Windows build adjustment code from @novemus  (see his magnificient code merge request https://github.com/savoirfairelinux/opendht/pull/716)
- Determined full-static and dynamic link libraries for  Windows if built using MSYS2 or MSVC
- Built without restinio and llhttp [TODO]
- Set for automatic fetching and population from the network the following libraries for MSYS2:
    - `GnuTLS` 3.8.4 for 64-bit Windows (tested for dynamic and static builds)
        - Auto name adjustments for `GnuTLS`
    - `GnuTLS` 3.8.9 for 32-bit Windows (untested)
    - `cJSON` as a dependency for msgpack
    - `msgpack` 6.1.1
    - `jsoncpp` 1.9.5
    - `libfmt` 11.1.2
    - `libasio` 1.30.2
- for MSVC:
    - `cJSON` (used by msgpack in dynamic builds)
    - `jsoncpp` 1.9.5
    - `libfmt` 11.1.2
- Various build scenario fixes for MSVC dynamic builds
- Added modular codebase for `infohash.cpp` to avoid header treatment mismatches in MSVC when compiling DLL
- Added convenience prints in the build scenario
- Fixed the inconvenient usage of `pkgconfig` in Windows, the link scenarios were adjusted accordingly
- Fixed hardcoded `find_library` calls
- Added standard CMake handler for Boost library using `BOOST_ROOT` variable, including environmental definition
- Fixed libreadline linkage on POSIX when the default link point is turned off for `OPENDHT_TOOLS` (Mint Linux)
- Added versioned handler for CMP0167 policy (from CMake 3.30)
- Code fixes for compatibility headers in Windows (`crypto.cpp`, `network_utils.h`, `utils.h`)
- Urgent fix for improper redefinition of `_WINNT_VERSION` for Windows 10/11 (https://github.com/savoirfairelinux/opendht/pull/716#issuecomment-2793129052)
- Compile definitions fixed for `opendht_EXPORTS` and `OPENDHT_BUILD` in build scenario for `OPENDHT_TOOLS`
- Urgent fix for GnuTLS version handler in the code (type redefinition in version 3.8.7 and later) https://github.com/savoirfairelinux/opendht/pull/716#issuecomment-2792592951
- Various fixes for proper dynamic linking in Windows
[...]

PR #758, opened 2025-06-03
Windows ci

aberaud:


twdragon:
@aberaud would you like to use the fixed version libraries populated in #750 also for this build?

twdragon:
> If that makes sense, I would prefer dependencies to be managed by vcpkg instead of cmake if possible. Llhttp (on unix) is an exception at this moment because it's not yet packaged on some platforms.

It is perfectly possible, it just requires removing or putting the fetcher under option. Do you want me to push this change?

TBH, what I really dislike now from the build pipeline is `WIN32_DEP_DIR`.

aberaud:
> > If that makes sense, I would prefer dependencies to be managed by vcpkg instead of cmake if possible. Llhttp (on unix) is an exception at this moment because it's not yet packaged on some platforms.
> 
> It is perfectly possible, it just requires removing or putting the fetcher under option. Do you want me to push this change?

Yes, all dependencies seem available on vcpkg, so at least as a first step we can make sure opendht builds with it on msvc.

> TBH, what I really dislike now from the build pipeline is `WIN32_DEP_DIR`.

Agreed :)

CMakeLists.txt
```diff
@@ -60,6 +60,15 @@ if (NOT HAVE_CXX_ATOMICS_WITHOUT_LIB
...
 
+if (WIN32)
+    if(CMAKE_SYSTEM_VERSION AND CMAKE_SYSTEM_VERSION VERSION_LESS_EQUAL "6.3")
+        set(WINVER 0x0603) # Minimum supported is Windows 8.1
+    else()
+        set(WINVER 0x0A00) # Windows 10/11 by default
+    endif()
+    add_definitions(-DWINVER=${WINVER} -D_WIN32_WINNT=${WINVER})
+endif()
+
```
twdragon, line 71: Do we really need this so explicitly?

The macro I used here: https://github.com/savoirfairelinux/opendht/pull/750/files#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR72-R89 does exactly this to avoid some problems on Windows installations where MSVC does not recognise the system version correctly. This leads to problems with `inet_pton` and other functions in case the real system version and `_WIN32_WINNT` point to different parts of code in `libgnutls` and WINAPI network stack.

CMakeLists.txt
```diff
@@ -155,25 +165,93 @@ if (NOT MSVC)
...
+            if (unofficial-readline-win32_FOUND)
+                set(readline_LIBRARIES unofficial::readline-win32::readline) 
+            else()
+                set (readline_LIBRARIES "readline")
+                message(WARNING "unofficial-readline-win32 not found, trying to use system readline")
+            endif()
+        endif ()
+    else()
+        # Traditional MSVC dependency path
+        set (WIN32_DEP_DIR ${PROJECT_SOURCE_DIR}/../)
```
twdragon, line 235: Could we remove `WIN32_DEP_DIR` completely after installation of all the packages in `vcpkg` and called `find_package`? It works normally in #750

CMakeLists.txt
```diff
@@ -155,25 +165,93 @@ if (NOT MSVC)
...
+            find_package(unofficial-readline-win32 CONFIG)
+            if (unofficial-readline-win32_FOUND)
+                set(readline_LIBRARIES unofficial::readline-win32::readline) 
+            else()
+                set (readline_LIBRARIES "readline")
+                message(WARNING "unofficial-readline-win32 not found, trying to use system readline")
+            endif()
+        endif ()
+    else()
+        # Traditional MSVC dependency path
```
twdragon, line 234: It is also better to set `/utf-8` and `/EHsc` to avoid any problems with implicitly defined (e.a, Cyrillic) Unicode strings

src/infohash.cpp
twdragon, line 1: Did you verify that deleting this traslation unit does not produce the DLL linking issue in MSVC?

CMakeLists.txt
aberaud: I saw what you did, and considered multiple strategies to define `_WIN32_WINNT`, including the stackoverflow solution.

The main rationale behind this strategy is that today, the only Windows API level still supported by Microsoft is essentially `0x0A00` (Windows 10/11). Microsoft still offers some commercial support for older versions, and some users are still on older Windows versions, so we would still support building for `0x0603` (Windows 8.1) when explicitly configured using `CMAKE_SYSTEM_VERSION`.

Is that reasonable? I would like to understand better the potential issues where MSVC does not recognize the system version correctly.

CMakeLists.txt
aberaud: Yes :)
The scope of this PR was to make the CI pipeline work, which is already huge.
I suggest further improvements can be brought by https://github.com/savoirfairelinux/opendht/pull/750 after that.

aberaud:
> @aberaud I set up an experiment with `vcpkg`, and I see that if you have all the packages installed system-wide, the `CMAKE_PREFIX_PATH` variable may be set to an empty string when the manifest is used. I may suggest not using `vcpkg.json` and concentrating on system-wide installed packages. I think it may also simplify the build pipeline. What do you think?
> 
> **UPDATE**: This behaviour is due to the `CMakePresets.json` file that overrides `CMAKE_PREFIX_PATH` from `vcpkg`, and also it may conflict with `CMakeSettings.json`. TBH, I don't know what to do with it unless we don't delete `CMakePresets.json` in addition to #761 .

I added a bunch of new presets, including "msvc" which doesn't set much except using VS2022. Would that work for your usecase ? It's also possible to not use presets, or to use userPresets.

twdragon:
> I added a bunch of new presets, including "msvc" which doesn't set much except using VS2022. Would that work for your usecase ? It's also possible to not use presets, or to use userPresets.

I tried the approach  but it is still zeroing out `CMAKE_PREFIX_PATH`, encountered the same behavior as in the past. In the context of the pipeline that includes CMake and Visual Studio, or either MSYS2, this method appears to be non-portable. I prefer using `userPresets.json`, and let the user write the proper generator name by himself. Also I would prefer this preset-less or with very minimal presetting approach for the whole project, as we are leaving the possibility to the user to build easily his own pipeline without patching anything. What do you think_?

twdragon:
I prepared an experimental MSI package for the library and tools built in SHARED mode in MSYS2: https://github.com/savoirfairelinux/opendht/pull/750#issuecomment-2962394293

twdragon:
@aberaud possible regression while building in STATIC mode on MSVC:

```
jsoncpp.lib(json_writer.cpp.obj) : error LNK2038: mismatch detected for 'RuntimeLibrary': value 'MT_StaticRelease' doesn't match value 'MD_DynamicRelease' i
n dhtscanner.obj [C:\Users\andrey.vukolov\src\opendht\build\tools\dhtscanner.vcxproj]
jsoncpp.lib(json_value.cpp.obj) : error LNK2038: mismatch detected for 'RuntimeLibrary': value 'MT_StaticRelease' doesn't match value 'MD_DynamicRelease' in
 dhtscanner.obj [C:\Users\andrey.vukolov\src\opendht\build\tools\dhtscanner.vcxproj]
```

I will try to check what is wrong with `OPENDHT_STATIC`

twdragon:
@aberaud the workaround for `MSVC_RUNTIME_LIBRARY` problem was added as a separate PR #762

PR #761, opened 2025-06-09
CMakePresets.json - remove UNIX preset for convenience

twdragon:
It is better to remove UNIX preset here, for the most users who prefer GNU Make generator


twdragon:
Closed in favour of #758

PR #762, opened 2025-06-12
Workaround for MSVC runtime mismatch on STATIC builds

twdragon:
- Added CMP0091 policy handler.
- Added generator expression for `MSVC_RUNTIME_LIBRARY` property to avoid "Runtime Mismatch" linker error on MSVC STATIC builds.


twdragon:
The only remaining warning now:

```
LINK : warning LNK4217: Символ "gnutls_free", определенный в "gnutls.lib(mem.obj)", imported "libopendht.lib(crypto.obj)" in ""private: class std::vector<unsigned char,class std::allocator<unsigned char> > __cdecl dht::crypto::PrivateKey::decryptBloc(unsigned char const *,unsigned __int64)const " (?decryptBloc@PrivateKey@crypto@dht@@AEBA?AV?$vector@EV?$allocator@E@std@@@std@@PEBE_K@Z)". 
    dhtscanner.vcxproj -> C:\Users\andrey.vukolov\src\opendht\build-vs\x64-Release-Static\tools\Release\dhtscanner.exe
    Building Custom Rule C:/Users/andrey.vukolov/src/opendht/CMakeLists.txt
```

aberaud:
Thanks, not sure where the warning comes from

twdragon:
It seems we have `__declspec(dllimport)` declaration on one of the function `dht::crypto::PrivateKey::decryptBloc` that calls implicitly `gnutls_free()`, but it does not bring breakouts, as I see

aberaud (approved):

PR #716, opened 2024-07-09
Windows build adjustment

novemus:
Apart from the cmake script adjustment, there are some minor code changes to fix dll export.

Steps to build opendht with Visual Studio 17 2022.

1. Install vcpkg and opendht dependencies (it takes a long time)
```console
git clone https://github.com/microsoft/vcpkg.git
cd vcpkg && bootstrap-vcpkg.bat
vcpkg install shiftmedia-libgnutls:x64-windows argon2:x64-windows asio:x64-windows fmt:x64-windows jsoncpp:x64-windows llhttp:x64-windows msgpack:x64-windows nettle:x64-windows openssl:x64-windows restinio:x64-windows
```
2. Build the project
```console
cd <opendht_root>
mkdir build && cd build
set PKG_CONFIG_PATH=<path_to_vcpkg_root>/installed/x64-windows/lib/pkgconfig
set CMAKE_PREFIX_PATH=<path_to_vcpkg_root>/installed/x64-windows/share
"C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Auxiliary\Build\vcvars64.bat"
cmake -S .. -G "Visual Studio 17 2022" -T host=x64 -A x64
cmake --build . --config Release --target ALL_BUILD
```

PR #763, opened 2025-06-13
add initial cpack support

aberaud:


twdragon:
@aberaud excellent initiative! Just a bit in advance: do you plan to use some parts of the pipeline from #750 and #716? I need to know this to plan my own activity on Windows related to OpenDHT, so I need a bit of clarity on the exact CMake pipeline

PR #767, opened 2025-06-17
Workaround for MSVC runtime mismatch on STATIC builds (for tests)

twdragon:
Added generator expression for `MSVC_RUNTIME_LIBRARY` property to avoid "Runtime Mismatch" linker error on MSVC STATIC builds for `cppunit` test runner executable.


aberaud:
should be fixed by https://github.com/savoirfairelinux/opendht/pull/768

PR #864, opened 2026-06-01
Fix UnifiedPush not working

vx32:
Two root causes:

1. The topic sent by the Android client is encoded with URL-safe base64. Add a separate auto-decoding function instead of modifying the original one, preserving symmetry between decode and encode.

2. Request's parameter is a string_view; in the low-level async handling it outlives its source and dangles. Make each class that uses the string_view hold its own owning string instead.

### DIFF
--- a/include/opendht/http.h
+++ b/include/opendht/http.h
@@ -344,8 +344,8 @@ private:
 
     dht::crypto::Identity client_identity_;
     std::shared_ptr<dht::crypto::Certificate> server_ca_;
-    std::string service_;
     std::string host_;
+    std::string service_;
 
     unsigned int id_;
     static std::atomic_uint ids_;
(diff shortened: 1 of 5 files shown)


aberaud:
@vx32 Thanks a lot for the contribution

I didn't like the URL "reconstruction" form this PR, but still merged most of this PR logic in master. Please let me know it that fixes the issue.

vx32:
> @vx32 Thanks a lot for the contribution
> 
> I didn't like the URL "reconstruction" form this PR, but still merged most of this PR logic in master. Please let me know it that fixes the issue.

`Push requests in the last 6.05 min: [Android: 0 high priority, 0 normal priority], [iOS: 0 high priority, 0 normal priority], [Unified: 12 high priority, 1 normal priority]
Yes, I've confirmed the issue is fixed. 
Your version of the change is more clean.  Thanks review.

PR #888, opened 2026-08-02
net: harden network message parsing and reassembly

cyrilleberaud:
## Problem

Malformed field-compressed responses can trigger a division by zero when their field selection is empty. Fragmented values also assume that the header arrives first and track in-progress messages only by transaction ID, so UDP reordering drops data and equal transaction IDs from different peers collide.

## Change

- reject empty field selections before computing the value count
- key partial messages by source address and transaction ID
- buffer fragments that arrive before their header
- bound pre-header bytes and packet count
- preserve rate limiting without charging reordered requests twice
- add regressions for malformed fields, reordered fragments, source collisions, empty fragment data and large values

## Tests

```
cmake --build build --target test_parsedmessage test_networkengine -j2
ctest --test-dir build --output-on-failure -R "test_(parsedmessage|networkengine)"
```

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>

### DIFF
--- a/include/opendht/network_engine.h
+++ b/include/opendht/network_engine.h
@@ -489,6 +489,8 @@ private:
     static constexpr size_t MTU {1280};
     static constexpr size_t MAX_PACKET_VALUE_SIZE {600};
     static constexpr size_t MAX_MESSAGE_VALUE_SIZE {56 * 1024};
+    static constexpr size_t RX_MAX_EARLY_BYTES {MAX_MESSAGE_VALUE_SIZE + MAX_VALUE_SIZE + 32};
+    static constexpr size_t RX_MAX_EARLY_PARTS {RX_MAX_EARLY_BYTES / MTU + 2};
 
     void process(std::unique_ptr<ParsedMessage>&&, const SockAddr& from);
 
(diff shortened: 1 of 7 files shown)


cyrilleberaud:
Note on CI: the \`Ubuntu/GCC minimal build (23)\` failure (\`4 - test_dhtrunner (Failed)\`) is pre-existing on \`master\` and unrelated to this change.

The same job fails identically on the base commit 56a60fb (run for \`master\` on 2026-07-30): \`88% tests passed, 1 tests failed out of 8\` / \`4 - test_dhtrunner (Failed)\`.

All other checks, including the new \`test_parsedmessage\` and \`test_networkengine\` regressions added here, pass.

cyrilleberaud:
Pushed a third commit to this branch: `net: reject malformed error and request fields`.

Continuing the audit of the message parser turned up two more remotely reachable defects in `ParsedMessage::msgpack_unpack()`:

1. **Null dereference on an empty error array.** `error_code = parsed.e->via.array.ptr[0].as<uint16_t>()` reads index 0 after checking only the type, never the size. An unauthenticated UDP packet carrying `"e":[]` crashes the process. The added regression segfaults on the unpatched tree (`Erreur de segmentation (core dumped)`, exit 139) and passes with the fix.

2. **Union type confusion on error-only messages.** `auto& req = parsed.a ? ... : *parsed.e;` made `req` alias the error array when a message had `e` but no `a`/`r`/`u`. The following loop then read `req.via.map.size` / `req.via.map.ptr[i]`, reinterpreting an `object*` array as an `object_kv*` map and reading roughly twice past its allocation. A request field that was not a map was never type-checked either.

The fix validates the error array size, restricts request parsing to `a`/`r`/`u`, and requires those to be maps.

`test_parsedmessage` now runs 27 tests, all green.

PR #889, opened 2026-08-02
value/dht: fix projected query deduplication and filtering

cyrilleberaud:
While porting the DHT logic I hit a divergence in projected-query deduplication.

## Problem

`FieldValueIndex::containedIn()` looks up each field key in the other index but never compares the resulting `FieldValue`:

```cpp
for (const auto& field : index) {
    auto other_field = other.index.find(field.first);
    if (other_field == other.index.end())
        return false;
}
```

The documented contract in `include/opendht/value.h` is *"Tells if all the fields of this are contained in the other FieldValueIndex **with the same value**"*, so the value comparison is simply missing.

## Impact

`Dht::query()` uses `containedIn()` to deduplicate projected results (`src/dht.cpp:1240-1254`). Every result of a given query shares the same projection schema, so any two results compared as mutually contained:

- the first result suppressed all subsequent ones;
- the `lesser` branch additionally erased the already reported result.

A query selecting a field therefore reported a single value no matter how many actually matched.

## Fix

Compare the `FieldValue` too. `FieldValue::operator==` already exists and handles every field kind.

## Test

Added `ValueTester::testFieldValueIndexContainedIn`, covering:

- two indexes projecting `Id` with different ids — neither contains the other;
- two equal indexes — mutually contained;
- a narrower projection contained in a wider one, and not contained when the value differs.

The test fails on `master` (`assertion failed - Expression: not first_index.containedIn(second_index)`) and passes with this change. The full local `ctest` suite passes (8/8).

Reviewers: @aberaud, plus francois-simon and virtual-reviewer per the daemon review policy (neither is assignable on GitHub).

### DIFF
--- a/include/opendht/value.h
+++ b/include/opendht/value.h
@@ -1032,8 +1032,22 @@ struct OPENDHT_PUBLIC Query
 
     /**
      * Tell if the query is satisfied by another query.
+     *
+     * The answer to `q` contains every value this query asks for, but it may
+     * contain more: the caller is responsible for applying this query's own
+     * filters to the values it receives.
      */
     bool isSatisfiedBy(const Query& q) const;
+
+    /**
+     * Tell if the projected answer to another query can be reported for this
+     * query without any further filtering.
+     *
+     * A field-projected answer only carries the fields selected by `q`, so a
+     * narrower filter can no longer be evaluated on it. Both queries must
+     * therefore apply exactly the same filters.
+     */
+    bool isProjectionSatisfiedBy(const Query& q) const;
 
     template<typename Packer>
     void msgpack_pack(Packer& pk) const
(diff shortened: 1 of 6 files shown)


cyrilleberaud:
Pushed a second, closely related commit: `dht: do not report unfiltered answers to filtered queries`.

`Query::isSatisfiedBy()` treats an unfiltered answer as sufficient for a filtered query. That holds for a vanilla get, because `onGetValuesDone()` applies the operation filter to the values before invoking `get_cb`.

It does not hold for a **projected** query. The `query_cb` branch forwards `a.fields` straight to the callback with no filtering, and those fields were selected for the *other* query — the filter may target a field that was not selected at all, so it cannot even be evaluated locally. On top of that, `searchNodeGetDone()` cancels the filtered query own request, so it never obtains a correctly filtered answer.

Concretely, two concurrent queries on the same key:

- `SELECT id`
- `SELECT id WHERE value_type=1`

made the second callback report identifiers whose type is not `1`.

Fix: add `Query::isProjectionSatisfiedBy()`, which additionally requires both queries to apply the same filters, and use it for operations carrying a query callback (`Dht::Get::isSatisfiedBy()`). Vanilla gets keep the existing, and correct, broader reuse.

Covered by `ValueTester::testProjectionSatisfiedBy`. Full local `ctest` passes (8/8).

PR #891, opened 2026-08-02
crypto: generate positive certificate serial numbers

cyrilleberaud:
Roughly half of the certificates generated by OpenDHT carry a **negative** serial number, which RFC 5280 forbids.

### Cause

`setRandomSerial()` draws a positive `int64_t`, then hands the native memory representation of that integer to `gnutls_x509_crt_set_serial()`:

```cpp
std::uniform_int_distribution<int64_t> dist {1};
int64_t cert_serial = dist(rdev);
gnutls_x509_crt_set_serial(cert, &cert_serial, sizeof(cert_serial));
```

GnuTLS expects that buffer in network byte order. On a little-endian host the least significant byte is passed first, so the leading DER byte is an arbitrary byte of the drawn value rather than its most significant one. When that byte has its high bit set, the DER INTEGER encodes a negative number.

Two consequences:

- RFC 5280 §4.1.2.2 states the serial number MUST be a positive integer. Negative serials are flagged by certificate linters and rejected by some X.509 implementations.
- The serial actually stored bears no relation to the number that was drawn, since the byte order is reversed.

### Reproduction

```cpp
int negative = 0;
for (int i = 0; i < 20; i++) {
    auto id = dht::crypto::generateIdentity("T", {}, 2048, true);
    auto serial = id.second->getSerialNumber();
    if (!serial.empty() && (serial[0] & 0x80)) negative++;
}
printf("negative=%d/20\n", negative);   // negative=11/20
```

`openssl x509 -noout -serial -text` on such a certificate reports e.g.
`Serial Number: -2663978395622007703 (-0x24f85a3c885d6397)`.

### Fix

Write the serial most significant byte first and clear the sign bit.

### Test

`CryptoTester::testGeneratedSerialNumberIsPositive` generates 16 identities and asserts the leading byte never has its high bit set. Verified failing before the change (`assertion failed - Expression: (serial.front() & 0x80) == 0`) and passing after it, alongside the 13 pre-existing crypto tests.

### DIFF
--- a/src/crypto.cpp
+++ b/src/crypto.cpp
@@ -17,6 +17,7 @@ extern "C" {
 #include <argon2.h>
 }
 
+#include <array>
 #include <random>
 #include <sstream>
 #include <fstream>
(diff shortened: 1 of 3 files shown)


aberaud:
fixed by https://github.com/savoirfairelinux/opendht/commit/cf64fa454c5b4ebb68aa2365f1569082e8c338d1

PR #899, opened 2026-08-03
peer discovery: use DNS-SD (rfc6763)

aberaud:


src/peer_discovery/mdns_transport.cpp
```diff
@@ -0,0 +1,455 @@
...
+        return address.multicast and not address.loopback and not address.address.is_unspecified()
+               and not address.address.is_multicast();
+    });
+
+    for (const auto& address : addresses) {
+        if (address.id == 0 or not address.multicast or (hasNonLoopback and address.loopback)
+            or address.address.is_unspecified() or address.address.is_multicast())
+            continue;
+
+        auto scope = std::find_if(result.begin(), result.end(), [&](const NetworkInterface& interface) {
```
Allium-altaicum, line 129: mingw defines "interface".
```
#define interface struct
```
So you can't use this name.🙃

PR #901, opened 2026-09-01
proxy_client: send push subscriptions deferred at token time

cyrilleberaud:
## Problem

`setPushNotificationToken()` stores the token and returns early when the proxy connection is not established yet, without ever scheduling the subscriptions it skipped:

```cpp
deviceKey_ = token;
if (!(statusIpv4_ == NodeStatus::Connected || statusIpv6_ == NodeStatus::Connected))
    return;                       // subscriptions are owed, and forgotten
```

Nothing retries them afterwards. `restartListeners()` only resubscribes listeners whose `opstate` is not `ok`, so healthy listeners keep the subscription they were created with and the server is never told the session id of the current run:

```cpp
for (auto& search : searches_)
    for (auto& listener : search.second.listeners)
        if (!listener.second.opstate->ok)
            resubscribe(search.first, listener.first, listener.second);
```

Clients set the push token while the account is starting, before the first proxy reply arrives, so this is the common path rather than a corner case.

## Consequence

The server keeps the listener registered under the session of a previous run and stamps every notification with that stale id. The client compares it against `pushSessionId_`, which is drawn once per `DhtProxyClient`, and discards the notification:

```cpp
if (sessionId != notification.end() and sessionId->second != pushSessionId_)
    return PushNotificationResult::IgnoredWrongSession;
```

The device silently stops receiving anything that depends on push: incoming calls, messages, and the diagnostic ping.

## Evidence

Measured on an Android device (Jami 20260807-01, OpenDHT 4.3.1 client, 4.2.0 proxy):

- **83 of 83** push notifications received were rejected with `IgnoredWrongSession`.
- Reproduced with the app in foreground and in background, screen locked and unlocked, battery optimised and whitelisted, and across full application restarts.
- The DHT side was verified healthy throughout: the device announcement was published, the caller's `PeerConnectionRequest` was present on the DHT at the expected key, and the push was delivered by the correct proxy at high priority.
- Restarting the proxy instance cleared the stale listeners. After that the device received **no push at all**, confirming it never subscribes for push with its current session.

## Fix

Remember the skipped subscriptions in `pushSubscriptionPending_` and send them from `restartListeners()` once the connection comes up.

`searchLock_` is already held at that point, so the loop is inlined rather than calling `resubscribeAll()`, which would deadlock. In `setPushNotificationToken()`, `lockCurrentProxyInfos_` is released before `searchLock_` is acquired, keeping the lock order consistent with `restartListeners()`.

Behaviour is unchanged when the connection is already established.

### DIFF
--- a/include/opendht/dht_proxy_client.h
+++ b/include/opendht/dht_proxy_client.h
@@ -368,6 +368,14 @@ private:
     std::string pushClientId_;
     std::string pushSessionId_;
 
+    /**
+     * Set when setPushNotificationToken() could not send the subscriptions
+     * because the proxy connection was not established yet. Consumed by
+     * restartListeners() once the connection comes up.
+     * Guarded by lockCurrentProxyInfos_.
+     */
+    bool pushSubscriptionPending_ {false};
+
     mutable std::mutex lockCurrentProxyInfos_;
     NodeStatus statusIpv4_ {NodeStatus::Disconnected};
     NodeStatus statusIpv6_ {NodeStatus::Disconnected};
@@ -439,6 +447,13 @@ private:
      * @param token
      */
     void resubscribe(const InfoHash& key, const size_t token, Listener& listener);
+
+#ifdef OPENDHT_PUSH_NOTIFICATIONS
+    /**
+     * Resubscribe every listener so the server learns the current push session.
+     */
+    void resubscribeAll();
+#endif
 
     /**
      * If we want to use push notifications by default.
--- a/src/dht_proxy_client.cpp
+++ b/src/dht_proxy_client.cpp
@@ -1205,6 +1205,25 @@ DhtProxyClient::restartListeners(const asio::error_code& ec)
         }
     }
     if (not deviceKey_.empty()) {
+        bool pending = false;
+        {
+            std::lock_guard l(lockCurrentProxyInfos_);
+            pending = pushSubscriptionPending_;
+            pushSubscriptionPending_ = false;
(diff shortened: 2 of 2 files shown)


cyrilleberaud:
Correction to my own analysis, after checking the objection raised in review.

The premise of this change is wrong. I claimed the subscriptions are never sent when the token is applied before the proxy connection is up. That is not what happens:

- the push token is passed to the **constructor**, not through this function:

```cpp
, pushSessionId_(getRandomSessionId())
, deviceKey_(pushToken)
```

- `setPushNotificationToken()` is only reached when the token **changes**. On the daemon side, `Account::setPushNotificationConfig()` returns `changed`, and the caller only forwards to the DHT runner when that is true. A normal account restart keeps the same FCM token, so this function is not called at all in the scenario I described;

- since `deviceKey_` is already set at construction, every listener is sent as a `subscribe`, which carries `session_id` through `getPushRequest()`. The session **is** communicated.

So this patch modifies a path that is not exercised on wake-up, and the mechanism I described does not hold. The reviewer was right.

I also ran an A/B on a physical device, patched build against an unpatched build from the same tree, same account, same protocol: five wake-up trials each, first push accepted in 5/5 on both sides. No measurable difference.

Two related changes are open: #903, whose justification leaned on the same analysis and should be judged on its own merits, and #904, which addresses a gap that is visible directly in the code rather than inferred.

Happy for this one to be closed. Apologies for the wasted review time.

